PR #22, opened 2024-08-03
Update const.py

ChristophCaina:
Added "de" Domain to const.py


ChristophCaina:
closed, as it seems not to be required for "de" URL...

chemelli74:
All countries that have a country code equal to the Amazon domain (de,es,it,...) are all automatically handled.
We need to document only exceptions.

PR #36, opened 2024-10-08
feat: updated output 

jeeftor:
- Change to html writing to write to the ./out directory and write either json or html based content-type
- To support this we also needed to add a fix for parsing empty JSON data


chemelli74:
Duplicate #43

PR #53, opened 2024-10-09
chore(library_test): add colorlog

chemelli74:


jeeftor:
I dont care which lib we use as long as we get color :)

PR #54, opened 2024-10-09
chore: optimize and fix debug files saving

chemelli74:


jeeftor:
Something to consider: 

https://docs.python.org/3.12/library/mimetypes.html



```python
import mimetypes

# Get the default extension for a MIME type
ext = mimetypes.guess_extension('text/plain')
print(ext)  # Typically outputs '.txt'

# Get the MIME type for a file extension
mime_type, _ = mimetypes.guess_type('example.jpg')
print(mime_type)  # Outputs 'image/jpeg'

# Get all extensions associated with a MIME type
exts = mimetypes.guess_all_extensions('application/javascript')
print(exts)  # Typically outputs ['.js', '.jsm']
```


Counter argument: We probably only have 2 types to deal with (.json) and (.html)

chemelli74:
We can add the dependency just for  development

Petro31 (commented):
Looks good to me.  I don't see anything that stands out.

PR #102, opened 2024-12-09
fix!: use ISO 3166 standard for country codes

chemelli74:
Now library aders to ISO 3166 standard.

As a consequence, United Kingdom users should now use "**`GB`**" and no more "`UK`".


chemelli74:
Now library aders to ISO 3166 standard.
As a consequence, United Kingdom users should now use "**`GB`**" and no more "`UK`".

PR #182, opened 2025-05-26
feat: add get_model_details() method

chemelli74:


src/aioamazondevices/api.py
```diff
@@ -629,6 +630,21 @@ async def auth_check_status(self) -> bool:
...
+        )
+        if not model_details:
+            _LOGGER.warning(
+                "Unknown device type '%s' for %s: please read https://github.com/chemelli74/aioamazondevices?tab=readme-ov-file#unknon-device-type",
+                device.device_type,
+                device.account_name,
+            )
+            return {}
+
+        return model_details
```
joostlek, line 646:
```suggestion
    async def get_model_details(self, device: AmazonDevice) -> dict[str, str] | None:
        """Return model datails."""
        model_details: dict[str, str] | None = DEVICE_TYPE_TO_MODEL.get(
            device.device_type
        )
        if not model_details:
            _LOGGER.warning(
                "Unknown device type '%s' for %s: please read https://github.com/chemelli74/aioamazondevices?tab=readme-ov-file#unknon-device-type",
                device.device_type,
                device.account_name,
            )

        return model_details
```

chemelli74: Typing is wrong because of <https://github.com/chemelli74/aioamazondevices/blob/main/src/aioamazondevices/const.py#L111>

PR #191, opened 2025-05-28
Update const.py

Mikey887:
added fix for .co.uk by specifying assoc handle


chemelli74:
Thx @Mikey887 for taking the time to investigate with me.

I'll wait for a few more countries to show up and see if it can be a common pattern, then I merge this change.

chemelli74:
Superseeded by #192

PR #205, opened 2025-05-31
fix: cookies for aiohttp

chemelli74:


src/aioamazondevices/api.py
```diff
@@ -262,6 +262,23 @@ def _client_session(self) -> None:
                     cookies=self._cookies,
                 )
 
+    async def _parse_cookies_from_headers(
+        self, headers: CIMultiDictProxy[str]
+    ) -> dict[str, str]:
+        """Parse cookies with a value from headers."""
+        cookies_with_value: dict[str, str] = {}
+
+        for value in headers.getall("Set-Cookie", []):
```
bdraco, line 271:
```suggestion
        for value in headers.getall("Set-Cookie", ()):
```

Tiny bit faster but doesn't really matter

chemelli74: Let's do it anyway !

PR #209, opened 2025-06-01
feat: add ability to launch skills

AzonInc:
This adds the ability to launch skills

### DIFF
--- a/library_test.py
+++ b/library_test.py
@@ -275,6 +275,13 @@ async def main() -> None:
     print(f"Text command on {device_single.account_name}")
     await api.call_alexa_text_command(device_single, "Set timer pasta 12 minute")
 
+    await wait_action_complete(10)
+
+    print("Launch 'MyTuner Radio' skill on ", device_cluster.account_name)
+    await api.call_alexa_skill(
+        device_cluster, "amzn1.ask.skill.94c477e7-61c0-43f5-b7d9-36d7498a4d04"
+    )
+
     await api.close()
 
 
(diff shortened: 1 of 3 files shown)


chemelli74:
Let's create a list of skills, so to make it easy for users.
This will also add the ability to validate input.

AzonInc:
I think it should be possible to use whatever skill id (we don't know all of the available skills around the world)
Or what do you mean by list?

library_test.py
```diff
@@ -273,6 +273,13 @@ async def main() -> None:
     print(f"Text command on {device_single.account_name}")
     await api.call_alexa_text_command(device_single, "Set timer pasta 12 minute")
 
+    await wait_action_complete(10)
+
+    print("Launch 'MyTuner Radio' on ", device_cluster.account_name)
```
chemelli74, line 278:
```suggestion
    print("Launch 'MyTuner Radio' skill on ", device_cluster.account_name)
```

src/aioamazondevices/api.py
```diff
@@ -906,6 +907,18 @@ async def _send_message(
...
+        elif message_type == AmazonSequenceType.LaunchSkill:
+            payload = {
+                **base_payload,
+                "targetDevice": {
+                    "deviceType": device.device_type,
+                    "deviceSerialNumber": device.serial_number,
+                },
+                "connectionRequest": {
+                    "uri": "connection://AMAZON.Launch/" + message_body,
+                    "input": {},
```
chemelli74, line 919:
```suggestion          
```

I don't think this is needed as empty placeholder

chemelli74 (changes requested):
Please rebase and check my comments

chemelli74:
@AzonInc do you mind completing the job so we can merge ?

AzonInc:
Sorry, I'm currently on holidays

AzonInc:
It's quite difficult to do it on phone but I think I managed

chemelli74:
Unfortunately you still have a commitlint error.

Please change the commit message from `Add feature to call alexa skills` to `feat: add feature to call Alexa skills`, so I can merge

PR #224, opened 2025-06-07
fix: handle requests response errors

chemelli74:


src/aioamazondevices/api.py
```diff
@@ -314,6 +319,20 @@ async def _session_request(
             content_type,
         )
 
+        if resp.status != HTTPStatus.OK:
```
bdraco, line 322: Can it return a 204?

bdraco (approved):

PR #227, opened 2025-06-08
Add ecobee Switch+

tronikos:
Also adding manufacturer which will be used by


tronikos:
Replaced by https://github.com/chemelli74/aioamazondevices/pull/231

PR #231, opened 2025-06-09
fix: Add ecobee Switch+

tronikos:
With https://github.com/home-assistant/core/pull/146333 that is added to the milestone for the next release all the properties are optional so no need to specify hw_version. Specify the manufacturer so that it correctly shows in the UI as ecobee Switch+ by ecobee Inc. This is exactly how the same device is shown via the HomeKit integration.

### DIFF
--- a/src/aioamazondevices/const.py
+++ b/src/aioamazondevices/const.py
@@ -169,8 +169,8 @@ DEVICE_TYPE_TO_MODEL: dict[str, dict[str, str | None]] = {
         "hw_version": "Gen11",
     },
     "AUPUQSVCVHXP0": {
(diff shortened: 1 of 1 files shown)


tronikos:
Replaced by https://github.com/chemelli74/aioamazondevices/pull/232

You need to fix your CI. It's too restrictive regarding commits. You should enable squash merging by default anyway and if you want commits to follow some pattern do that on PR titles which can easily change in the UI.

tronikos:
When using the web editor at github.dev you can't use pre-commit.
And you can't easily change commit subjects.
If you squash merge the individual commits don't matter anyway. Only the PR title matters.

PR #232, opened 2025-06-09
fix: add ecobee switch+

tronikos:
Specify the manufacturer so that it correctly shows in the UI as ecobee Switch+ by ecobee Inc. This is exactly how the same device is shown via the HomeKit integration.


chemelli74:
Please add `"hw_version": None` anyway, so it's clear it's not forgot; I also use a script to validate data and sort them that rely on all fields for the above reason.

tronikos:
Added `"hw_version": None`

PR #240, opened 2025-06-12
Add MX assoc_handle

lchavezcuu:


novasdream:
I believe that logic already exist on https://github.com/chemelli74/aioamazondevices/blob/359c1c41493e04a9cf9ef83934ef2c8126889d38/src/aioamazondevices/api.py#L112

lchavezcuu:
> Hi @lchavezcuu , thx for your effort but I bet you didn't test your change; did you ?
> 
> 
> 
> 

Hi, I did not test it, I will look for documentation on how to do it.

lchavezcuu:
> > Hi @lchavezcuu , thx for your effort but I bet you didn't test your change; did you ?
> 
> Hi, I did not test it, I will look for documentation on how to do it.

@chemelli74 I was able to run the test following the instructions in the readme, I was able to validate a notification is pushed to the echo I specified, I even heard some music and some sort of podcast I think, this seems to be working

I also run the same test with current `main` (without this config) and I was able to see an error in the console, so the settings Im adding are fixing the problem for MX

chemelli74:
Hi, nice job !

Please do another test with just the domain and without the specific assoc_handle.

Thanks

lchavezcuu:
> Hi, nice job !
> 
> Please do another test with just the domain and without the specific assoc_handle.
> 
> Thanks

@chemelli74 It did work, I removed the assoc_handle for `mx`

PR #242, opened 2025-06-12
fix: 5 new models and 2 name fixes

Ivo-tje:
Added and corrected some models


Ivo-tje:
> Hi @Ivo-tje , thx for reporting.
> 
> Can you please show me a referecence for the updated model name ? Seems there are different users reporting that with different names.
> 
> Note: please amend the commit message to be commitlint compliant.

I got a bunch of Alexa's, both echo dot gen4s with the same ID don't have the clock. The other one isn't a dot for sure as this is the high white pole model. The added once got the name as displayed in the amazon device page.

chemelli74:
> finally all checks green!

Not really unfortunately.

Also, as Sonos so far are not confirmed to be working I prefer not to add them yet.

Ivo-tje:
If Sonos is working or not doesn't matter, now they are listed as unknown and that would be fixed so you know at least which type of device it is.

chemelli74:
Superseded by #295

Issue #245, opened 2025-06-12
Unknown Device Types Fire 8 Pro, Fire Stick, App for PC


jamesonuk:
- A1RTAM01W29CUP - Is "Alexa App for PC` and shows up even though you can't actually use this any more.   I think these devices should actually be ignored
- ADVBD696BHNV5 - Fire TV Stick (Gen 1)
- A1TD5Z1R8IWBHA - Fire HD 8 (12th Gen)
- A15ERDAKK5HQQG - Sonos.   It is a Play 5 (first gen) but https://github.com/alandtse/alexa_media_player/blob/dev/custom_components/alexa_media/const.py and Amazon account just show it as Sonos so not sure if this covers more than one physical device type

jamesonuk:
> Can you confirm the 2 notify entities works fine please ?

Possibly another issue to raise but I can't actually get the notify entities to work on any of my devices....

```
action: notify.send_message
data:
  message: test
target:
  entity_id: notify.study_dot_announce_dot_announce
```

seems to say it has worked but nothing gets announced.   I did see a HTTP 401 in the debug logs but I need a little time to look properly

chemelli74:
Can you please attach the diagnostic of the Sonos device ?

jamesonuk:
> Can you please attach the diagnostic of the Sonos device ?

[alexa_devices-01JXHXJ39P0DZHY2C64KSW9C10-Playroom-7124c857c122364447f833862df58897.json](https://github.com/user-attachments/files/20709525/alexa_devices-01JXHXJ39P0DZHY2C64KSW9C10-Playroom-7124c857c122364447f833862df58897.json)

PR #246, opened 2025-06-12
Updated models

jamesonuk:
New models as per https://github.com/chemelli74/aioamazondevices/issues/245


src/aioamazondevices/const.py
```diff
@@ -255,4 +255,21 @@
         "model": "Echo Plus",
         "hw_version": "Gen2",
     },
+    "A1RTAM01W29CUP": {
+        "model": "Alexa App for PC",
```
chemelli74, line 259: WE want to ignore that as it cannot be used anymore

chemelli74 (changes requested):
Please review my comments and fix CI

PR #258, opened 2025-06-13
feat: ask for login password using getpass

ivanfmartinez:
This way the password are not stored in clear text on json neither user history file


library_test.py
```diff
@@ -150,13 +151,12 @@ async def main() -> None:
...
 
-    if not args.password:
-        print("You have to specify a password")
-        parser.print_help()
-        sys.exit(1)
-
     login_data_stored = read_from_file(args.login_data_file)
 
+    if not login_data_stored and not args.password:
+        print("You have to specify a password")
```
chemelli74, line 157:
```suggestion
        print(f"You have to specify credentials for {args.email}")
```

library_test.py
```diff
@@ -150,13 +151,12 @@ async def main() -> None:
...
+        args.password = getpass.getpass(f"enter the password for {args.email}: ")
```
chemelli74, line 158:
```suggestion
        args.password = getpass.getpass("Password: ")
```

chemelli74 (changes requested):
Thx for the contribution !

Just a small change to the prompt and we are good to go

chemelli74 (approved):

PR #265, opened 2025-06-14
fix: redact sensitive data from debug logs

chemelli74:


jamesonuk:
Couple more things you probably want to hide.
In the call to `device-preferences`  the array has
* `deviceAddress` (address as string)
* `deviceAddressModel` (address as object split into parts)
* `postalCode`

I also suspect these should be hidden too
* `deviceSerialNumber`  which I think is probably widespread
* `searchCustomerId`


Just putting these here whilst I was trying to sanitise logs to post

chemelli74:
> * `deviceAddress` (address as string)
> * `postalCode`

Thse are already listed and  part `deviceAddressModel` that is redacted.

> * `deviceSerialNumber`  which I think is probably widespread

This one is needed for troubleshooting unfortunately

> * `searchCustomerId`

Will add this one

Issue #272, opened 2025-06-15
Handle Amazon Rate Limits


jamesonuk:
I updated the library to 3.1.13 to get past the `serial_number` error you fixed in https://github.com/chemelli74/aioamazondevices/pull/264

However the integration still would not start with
```
File "/usr/local/lib/python3.13/site-packages/aioamazondevices/api.py",  
    line 700, in get_devices_data self._devices[dev_serial] = {key: data}
File "/usr/local/lib/python3.13/site-packages/aioamazondevices/api.py",   
    line 513, in _get_devices_ids network_detail = orjson.loads(json_data["networkDetail"]) 
KeyError: 'networkDetail'
```

doing a bit of debugging and I found that the call 
https://github.com/chemelli74/aioamazondevices/blob/main/src/aioamazondevices/api.py#L507-L510

is actually getting a HTTP 429 with the response `{'message': 'Too many requests. Try again in some time.'}` hence the `KeyError`

Think there are a couple of things that need to be added here.   First off is checking the HTTP response and handling errors.   Assuming the rate limit won't get exceeded going forward then I guess it will sort itself out eventually but would be a lot easier to debug if that was handled and logged.

This is partly handled in other places like 
https://github.com/chemelli74/aioamazondevices/blob/main/src/aioamazondevices/api.py#L765-L770
but not `_get_devices_ids` and `_get_sensor_states`    Having had the JSON output on a non HTTP 200 status in the debug logs would have got me to the rate limiting error a lot quicker

Second bit is about checking the keys exist before reading.   There are a few places where the code will error if the JSON is not in the right format (and being Amazon I would highly expect them to alter their structure at the drop of a hat).    Given the response if not under your control I think it would be prudent to validate it matches the expected structure and log where it doesn't

chemelli74:
> 2025-06-15 09:03:58.703 DEBUG (MainThread) [aioamazondevices] Response 429 for url https://alexa.amazon.co.uk/api/phoenix with content type: application/json

Error 429, Too many requests.
Will take care of it

jamesonuk:
> I'm quite sure there is another erro before those lines

before that you have what appears to be a working request that has returned data OK

```
2025-06-15 09:03:58.295 DEBUG (MainThread) [aioamazondevices] GET request: https://alexa.amazon.co.uk/api/bluetooth with payload None [json=False]
2025-06-15 09:03:58.661 DEBUG (MainThread) [aioamazondevices] Cookies from headers: {}
[...]

PR #273, opened 2025-06-15
Handle unexpected responses

jamesonuk:
Not the most elegant but would go towards addressing some of https://github.com/chemelli74/aioamazondevices/issues/272

Might need a better fix to handle rate limiting but this would avoid the errors

### DIFF
--- a/src/aioamazondevices/api.py
+++ b/src/aioamazondevices/api.py
@@ -55,6 +55,7 @@ from .exceptions import (
     CannotRegisterDevice,
     RequestFailed,
     WrongMethod,
+    UnexpectedResponse,
 )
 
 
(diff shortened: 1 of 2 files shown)


chemelli74:
Fixed via #274

jamesonuk:
> Fixed via #274

This does push the HTTP response back up and you get see a 429 response (although as per comment on PR#274 there is a typo which means `%s` gets output as a literal.   Would be a lot more user friendly to see the response message as 429 means something to me but I suspect won't mean much to others (where as `{'message': 'Too many requests. Try again in some time.'}` would)

Also `KeyError`  on the structure is still likely (https://github.com/home-assistant/core/issues/146777) and it would be useful to log the unexpected response for debugging still

PR #276, opened 2025-06-16
fix: optimize exception handling

chemelli74:


src/aioamazondevices/api.py
```diff
@@ -328,13 +329,17 @@ async def _session_request(
...
+        try:
+            resp = await self.session.request(
+                method,
+                URL(url, encoded=True),
+                data=input_data if not json_data else orjson.dumps(input_data),
+                cookies=_cookies,
+                headers=headers,
+            )
+        except (TimeoutError, ClientConnectorError) as exc:
+            raise CannotConnect("Connection error during %s", method) from exc
```
jamesonuk, line 341: this will output `%s` as a literal.    Use 
` raise CannotConnect("Connection error during {0}".format(method)) from exc`
or 
`raise CannotConnect("Connection error during %s" % method) from exc`
but the top one is preferred way to do it in Python now

src/aioamazondevices/api.py
```diff
@@ -457,12 +464,13 @@ async def _register_device(
...
         if resp.status != HTTPStatus.OK:
+            msg = resp_json["response"]["error"]["message"]
             _LOGGER.error(
                 "Cannot register device for %s: %s",
                 self._login_email,
-                resp_json["response"]["error"]["message"],
+                msg,
             )
-            raise CannotRegisterDevice(resp_json)
+            raise CannotRegisterDevice("%s: %s", HTTPStatus(resp.status).phrase, msg)
```
jamesonuk, line 473: Another one which will output `%s: %s` as part of message

Issue #279, opened 2025-06-16
typo when checking for 404 response ?


ivanfmartinez:
After checking my login problem I started getting the 404 response from amazon, and it does not login.

I found that this code is using "sigin" instead of "signin" is that correct ?

https://github.com/chemelli74/aioamazondevices/blob/8b527e029976e85975bb0d6ed10e8ce3d5cfdeec/src/aioamazondevices/api.py#L294

https://github.com/chemelli74/aioamazondevices/blob/8b527e029976e85975bb0d6ed10e8ce3d5cfdeec/src/aioamazondevices/api.py#L299

jamesonuk:
I am not entirely sure this is right through.    I deleted and tried to add and came across this error but simply fixing the typo didn't resolve it.

I tracked it back through and the first bit is that I was sent to `/ap/maplanding` which I am not sure if because I have a passkey setup for Amazon and I can't quite match how `aiohttp` is handling things to what I expect.

First off following the docs, `response.url.name` should only return the last part of the URL path (so for https://amazon.co.uk/ap/signin it should only return `signin` 

https://docs.aiohttp.org/en/stable/client_reference.html#aiohttp.ClientResponse.url -> https://yarl.aio-libs.org/en/stable/api/#yarl.URL.name

which is what I was seeing with just `maplanding` and hence I can't see how this code would work (but obviously is and does 😕 )
this path is handled explicitly in `alexapy` https://gitlab.com/search?search=mapland&nav_source=navbar&project_id=7993293&search_code=true&repository_ref=dev

I put in exceptions for `response.url.path` being either `/ap/signin` or `/ap/maplanding` and then it logged me in (albeit I just hit the 429 rate limiting issue again)

PR #281, opened 2025-06-16
fix: correct check for 404, and save response before checking for errors for easy debug

ivanfmartinez:
I started getting the 404 error and cannot login after the await inclusion.

I found what I think was typo and also the checking for url using response.url was not working.

Also moved the save response before checking.

If there was a typo as I asked on https://github.com/chemelli74/aioamazondevices/issues/279 we can remove the check for "/sigin" and keep only check for "/signin"

With this changce I was able to login again in my account that I cannot login on latest releases.

### DIFF
--- a/src/aioamazondevices/api.py
+++ b/src/aioamazondevices/api.py
@@ -295,8 +295,8 @@ class AmazonEchoApi:
         """Return true if error is due to /ap/signin endpoint."""
         # Endpoint 'ap/signin' replies with error 404
         # but reports the needed parameters anyway
-        return (
-            response.status == HTTPStatus.NOT_FOUND and ("/ap/signin" in url or "/ap/sigin" in url)
+        return response.status == HTTPStatus.NOT_FOUND and (
+            "/ap/signin" in url or "/ap/sigin" in url
         )
 
     async def _session_request(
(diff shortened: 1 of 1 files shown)


src/aioamazondevices/api.py
```diff
@@ -350,14 +350,21 @@ async def _session_request(
...
+
         if resp.status != HTTPStatus.OK:
             if resp.status in [
                 HTTPStatus.FORBIDDEN,
                 HTTPStatus.PROXY_AUTHENTICATION_REQUIRED,
                 HTTPStatus.UNAUTHORIZED,
             ]:
                 raise CannotAuthenticate(HTTPStatus(resp.status).phrase)
-            if not await self._ignore_ap_sigin_error(resp):
+            if not await self._ignore_ap_signin_error(resp, url):
```
jamesonuk, line 367: Given the function is only called from one place and we are now passing multiple parameters, wondering whether it would be more readable to simply check the `HttpStatus.NOT_FOUND` and that request was for `/ap/signin` here rather than in a function, we are already checking `HttpStatus` above

chemelli74:
Hi, I went for #285

ivanfmartinez:
> Hi, I went for #285

Did not worked for me. 

I have made prints and now the response.url are pointing to correct URL, and this history is empty.

Maybe amazon sometimes are redirecting and sometimes not, in this case history and request.url should be checked

chemelli74:
> Maybe amazon sometimes are redirecting and sometimes not, in this case history and request.url should be checked

Basically when it doesn't, then there is a login problem of some kind.
Will intestigate further.

PR #288, opened 2025-06-17
fix: update User-Agent

jamesonuk:
Update User-Agent (as per https://github.com/home-assistant/core/issues/146784#issuecomment-2974813319) 

hopefully will resolve 429 response in https://github.com/chemelli74/aioamazondevices/issues/272


chemelli74:
Thx for taking the time to test.

Should we also update `AMAZON_APP_VERSION` and `AMAZON_CLIENT_OS` to more recent values ?

jamesonuk:
> Should we also update `AMAZON_APP_VERSION` and `AMAZON_CLIENT_OS` to more recent values ?

I did but when I switched over from dev machine I forgot to do it...   two secs

PR #309, opened 2025-06-23
feat: queues API calls to avoid hitting the limit

### LINKED ISSUE #307: Handle Rate Limiting
I reported an issue with 429 (TOO MANY REQUESTS) responses which didn't actually appear to be rate limiting but rather blocking of calls by Amazon.

I am now seeing what appear to be real 429 responses.

Simple way to replicate it to create an announcement that targets multiple devices.    I have one announcing to three devices and it only takes one or two calls to trigger the error.

I have not had chance to look at https://github.com/chemelli74/aioamazondevices/pull/284 but there needs to be something to handle 429 responses and retry.   There is definitely no `Rety-After` header being returned (nor anything else particularly helpful)

Not sure if in this situation there is a way to combine the calls so if you are announcing to more than one device you can do this is in one call as opposed to one per device? (either way handling 429 responses is going to be needed)

chemelli74:
- Closes #307

### DIFF
--- a/src/aioamazondevices/api.py
+++ b/src/aioamazondevices/api.py
@@ -373,7 +373,9 @@ class AmazonEchoApi:
 
         for delay in [0, 1, 2, 5, 8, 12, 21]:
             if delay:
-                _LOGGER.debug("Sleeping for %s seconds before retrying API call", delay)
+                _LOGGER.info(
+                    "Sleeping for %s seconds before retrying API call to %s", delay, url
+                )
                 await asyncio.sleep(delay)
 
             try:
@@ -392,6 +394,7 @@ class AmazonEchoApi:
                 ]:
                     break
             except (TimeoutError, ClientConnectorError) as exc:
+                _LOGGER.warning("Connection error to %s: %s", url, repr(exc))
                 raise CannotConnect(f"Connection error during {method}") from exc
 
         if not self._csrf_cookie:
--- a/src/aioamazondevices/api.py
+++ b/src/aioamazondevices/api.py
@@ -371,6 +371,7 @@ class AmazonEchoApi:
         )
         self.session.cookie_jar.update_cookies(_cookies)
 
+        resp: ClientResponse | None = None
         for delay in [0, 1, 2, 5, 8, 12, 21]:
             if delay:
                 _LOGGER.info(
@@ -396,6 +397,10 @@ class AmazonEchoApi:
             except (TimeoutError, ClientConnectorError) as exc:
                 _LOGGER.warning("Connection error to %s: %s", url, repr(exc))
                 raise CannotConnect(f"Connection error during {method}") from exc
+
+        if resp is None:
+            _LOGGER.error("No response received from %s", url)
+            raise CannotConnect(f"No response received from {url}")
 
         if not self._csrf_cookie:
             self._csrf_cookie = resp.cookies.get(CSRF_COOKIE, Morsel()).value
--- a/src/aioamazondevices/api.py
+++ b/src/aioamazondevices/api.py
@@ -386,17 +386,19 @@ class AmazonEchoApi:
                     data=input_data if not json_data else orjson.dumps(input_data),
                     headers=headers,
                 )
-                # Retry with a delay only for specific HTTP status
-                # that can benefits of a back-off
-                if resp.status not in [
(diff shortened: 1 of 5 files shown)


jamesonuk (approved):
This is looking clean and neat and will resolve real rate limiting requests.
(like the ones I can replicate sending announcements to multiple devices)
Still going to see 429s where access is temporarily blocked but graphQL or possibly randomising app version might help here.

This will solve real rate limiting.

chemelli74:
@copilot there are conflicts in this PR, please resolve them

src/aioamazondevices/api.py
```diff
@@ -369,15 +370,29 @@ async def _session_request(
...
-                URL(url, encoded=True),
-                data=input_data if not json_data else orjson.dumps(input_data),
-                headers=headers,
-            )
-        except (TimeoutError, ClientConnectorError) as exc:
-            raise CannotConnect(f"Connection error during {method}") from exc
+
+        for delay in [0, 1, 2, 5, 8, 12, 21]:
+            if delay:
+                _LOGGER.debug("Sleeping for %s seconds before retrying API call", delay)
```
jamesonuk, line 376: I think it would be useful to log the url we are sleeping on.
This can get intertwined with other calls and it isn't always obviously what the sleep relates to.

Perhaps just 
```python
_LOGGER.debug("Sleeping for %s seconds before retrying API call to %s", delay, url)
```

src/aioamazondevices/api.py
```diff
@@ -369,15 +370,29 @@ async def _session_request(
...
+                    HTTPStatus.INTERNAL_SERVER_ERROR,
+                    HTTPStatus.SERVICE_UNAVAILABLE,
+                    HTTPStatus.TOO_MANY_REQUESTS,
+                ]:
+                    break
+            except (TimeoutError, ClientConnectorError) as exc:
+                raise CannotConnect(f"Connection error during {method}") from exc
 
         if not self._csrf_cookie:
             self._csrf_cookie = resp.cookies.get(CSRF_COOKIE, Morsel()).value
```
jamesonuk, line 398: Pylance is complaining that resp might be unbound here and below.
I think it would be safer to add an explicit `resp = None` assignment before for loop and add a check after to ensure it is now set

src/aioamazondevices/api.py
chemelli74: Good idea, done

src/aioamazondevices/api.py
```diff
@@ -369,15 +370,37 @@ async def _session_request(
...
-                data=input_data if not json_data else orjson.dumps(input_data),
-                headers=headers,
-            )
-        except (TimeoutError, ClientConnectorError) as exc:
-            raise CannotConnect(f"Connection error during {method}") from exc
+
+        resp: ClientResponse | None = None
+        for delay in [0, 1, 2, 5, 8, 12, 21]:
+            if delay:
+                _LOGGER.info(
```
jamesonuk, line 377: Just checking you want info logging here.    I would prefer this to be debug as for most users this will just be noise and it might happen more often than we think

src/aioamazondevices/api.py
```diff
@@ -369,15 +370,37 @@ async def _session_request(
...
+                )
+                # Retry with a delay only for specific HTTP status
+                # that can benefits of a back-off
+                if resp.status not in [
+                    HTTPStatus.INTERNAL_SERVER_ERROR,
+                    HTTPStatus.SERVICE_UNAVAILABLE,
+                    HTTPStatus.TOO_MANY_REQUESTS,
+                ]:
+                    break
+            except (TimeoutError, ClientConnectorError) as exc:
```
jamesonuk, line 397: Other tiny thing is whether the except block should move to be after the try and put the check on the response status out side of the try block?

src/aioamazondevices/api.py
chemelli74: I moved to info because at least everyone knows there is a delay in the call, otherwise at debug level the majority will miss it and complain.

src/aioamazondevices/api.py
chemelli74: Why ? What would be the benefit ?

jamesonuk (approved):
Totally happy with this and has been running fine on my system for a few hours now (triggering it with a few announcements to multiple speakers)

PR #316, opened 2025-06-24
fix: ignore devices earlier

jamesonuk:
move where devices get ignored earlier to avoid KeyErrors for incompatible devices

### DIFF
--- a/src/aioamazondevices/api.py
+++ b/src/aioamazondevices/api.py
@@ -723,15 +723,11 @@ class AmazonEchoApi:
 
         final_devices_list: dict[str, AmazonDevice] = {}
         for device in self._devices.values():
-            if NODE_DEVICES not in device:
-                _LOGGER.debug("%s not in %s", NODE_DEVICES, device)
+            # Remove stale, orphaned and virtual devices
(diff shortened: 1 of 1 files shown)


src/aioamazondevices/api.py
```diff
@@ -723,7 +723,15 @@ async def get_devices_data(
...
         for device in self._devices.values():
+            if NODE_DEVICES not in device:
+                _LOGGER.debug("%s not in %s", NODE_DEVICES, device)
+                continue
             devices_node = device[NODE_DEVICES]
+            if devices_node.get("deviceType") in DEVICE_TO_IGNORE:
+                _LOGGER.debug(
+                    "Skipping device type {%s}", devices_node.get("deviceType")
+                )
+                continue
```
chemelli74, line 734:
```suggestion
            # Remove stale, orphaned and virtual devices
            devices_node = device.get(NODE_DEVICES)
            if not devices_node or (devices_node.get("deviceType") in DEVICE_TO_IGNORE):
                continue
```

chemelli74 (changes requested):
An optimized version

PR #325, opened 2025-06-25
fix: add Israel domain settings

Pristavkin:
`python library_test.py ` works ok, but fails on another [429 rate limit issue](https://github.com/chemelli74/aioamazondevices/issues/307)

### DIFF
--- a/src/aioamazondevices/const.py
+++ b/src/aioamazondevices/const.py
@@ -48,6 +48,10 @@ DOMAIN_BY_ISO3166_COUNTRY = {
         "domain": "co.uk",
         "openid.assoc_handle": f"{DEFAULT_ASSOC_HANDLE}_uk",
     },
+    "il": {
+        "domain": "com",
+        "openid.assoc_handle": DEFAULT_ASSOC_HANDLE,
+    },
     "jp": {
         "domain": "co.jp",
     },


chemelli74:
Israel doesn't look like an available coutnry for Amazon, I bet your account is created in US.

If so you have to select US as country

Pristavkin:
> Israel doesn't look like an available coutnry for Amazon, I bet your account is created in US.
> 
> If so you have to select US as country

Hmm, I’m pretty sure my account has always been registered in Israel. I’m not exactly sure how to verify that, but here’s a screenshot of the device profile page:
image

Also, even if you're right, I found the same settings tuning for Argentina, so I’m not sure why it should behave different for Israel. 

And the last thing, it will be valuable for Home Assistance usability, as integration by default use country code from Home Assistance instance. So this issue should be ether clearly documented in the integration manual page, or adjusted in the code. Last one is more preferable from my standpoint.

chemelli74:
> > Israel doesn't look like an available coutnry for Amazon, I bet your account is created in US.
> > If so you have to select US as country
> 
> Hmm, I’m pretty sure my account has always been registered in Israel. I’m not exactly sure how to verify that,

Which is the login page you use from the browser ?

> Also, even if you're right, I found the same settings tuning for Argentina, so I’m not sure why it should behave different for Israel.

You are right, was a mistake adding that country as it gives the wrong idea that you can select Argentina. While instead you have to select US for amazon.com.


> And the last thing, it will be valuable for Home Assistant usability, as integration by default use country code from Home Assistant instance. 

That's already the case.

> So this issue should be ether clearly documented in the integration manual page, or adjusted in the code. Last one is more preferable from my standpoint.

I think it's clearly stated in the documentation: https://www.home-assistant.io/integrations/alexa_devices#country

Pristavkin:
> Feel free to write a PR to update the documentation

Nah, I’ll just fork the library, apply the necessary updates, and bind it on the Docker side.

Pristavkin:
Forked library, not relevant anymore

chemelli74:
> > > > And the last thing, it will be valuable for Home Assistant usability, as integration by default use country code from Home Assistant instance.
> > > 
> > > 
> > > That's already the case.
> > 
> > 
> > Yep, and that’s actually part of the issue. From a usability standpoint, it causes confusion, as user can legitimately select Israel (and even more, it's selected by default), but as a result will have just a vague 'Failed to connect' message.
> > So if Israel isn’t a valid option, it should excluded from the list of available countries (as a many other). But that would mean maintaining a list of valid countries supported by Amazon for this integration.
> 
> Will think about it, not sure which is the best approach on that.

After looking at the outcome of the integration in the last weeks, I think is worth to add Israel country settings to the const.py
as it's one of the few that needs special settings.

PR #331, opened 2025-06-26
chore: cleanup redundant output

jamesonuk:
Just cleanup redundant output which is already saved in  `_session_request` call preceeding this


chemelli74:
Thx!

PR #335, opened 2025-06-30
feat: check country setting

chemelli74:


jamesonuk (approved):
Tested and worked without issue.
Tried to setup with wrong country and this worked as expected

PR #339, opened 2025-07-01
chore: amend delay loop logic

jamesonuk:
there is an issue where when the /ap/signin page expectedly returns a HTTP 404 the loop is not interrupted so the URL is requested again leading to being unable to sign in.

Also once the delays have been exhausted the loop just ends with no exception being raised

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 ## v3.3.0-rc.1 (2025-07-01)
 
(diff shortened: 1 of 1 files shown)


chemelli74:
Thx for your help !!!

Please limit the changes to those mandatory for fixing the issue, thus leaning the method in place.

Rework, if really wanted, should be in a separate PR.

PR #348, opened 2025-07-05
fix: update model for Echo Show 5 gen1

alessio-papi:
Wrong label for  Echo Show 5 gen1


chemelli74:
Hi, are you really sure ?

Is plenty of sites reporting this code as "Echo Show 5".

alessio-papi:
Ciao Simone,

I'm pretty sure:

- I have that device (Echo Show 5 gen1)
-  double checked on other sites (ie alexa media player)
-  try to find Echo Show **3**, but I doesn't exit

so I should be 5 and not 3 :-)

jamesonuk:
I have fixed the commit message and pushed https://github.com/chemelli74/aioamazondevices/pull/363 with your commit

PR #352, opened 2025-07-06
fix: improve language settings

chemelli74:


jamesonuk (approved):
This works for me (but only when PR#353 and PR#356 are applied but that is issue in main not to do with this branch).

Worked with both clean and existing login data and ran three times each.

PR #353, opened 2025-07-06
fix: ignore sensors if not available

chemelli74:


jamesonuk (approved):
Tested in both HA and library.   No more failures due to 299 HTTP responses.
Warning is logged in HA.

Everything now working (except sensors, which aren't there as expected)

PR #356, opened 2025-07-06
chore: don't process devices if no data

jamesonuk:


chemelli74:
Similar solution included in #353

PR #359, opened 2025-07-08
chore: handle unknown message_type

jamesonuk:
Just linting picked up by Pylance.
ensures unhandled `message_type` is raised

### DIFF
--- a/src/aioamazondevices/api.py
+++ b/src/aioamazondevices/api.py
@@ -1017,6 +1017,8 @@ class AmazonEchoApi:
                 "skillId": "amzn1.ask.1p.tellalexa",
                 "text": message_body,
             }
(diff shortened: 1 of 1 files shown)


src/aioamazondevices/api.py
```diff
@@ -1017,6 +1017,8 @@ async def _send_message(
                 "skillId": "amzn1.ask.1p.tellalexa",
                 "text": message_body,
             }
+        else:
+            raise ValueError("Message type <{message_type}> is not recognised")
```
chemelli74, line 1021:
```suggestion
            raise ValueError(f"Message type <{message_type}> is not recognised")
```

chemelli74 (approved):

PR #367, opened 2025-07-13
feat: use CookieJar instead of SimpleCookie

chemelli74:


jamesonuk (approved):
ran two tests with no saved credentials and two with saved credentials and both ran through without issue.

PR #369, opened 2025-07-14
feat!: use external aiohttp ClientSession

chemelli74:


jamesonuk (approved):
Another two clean and two existing runs and everything worked fine

chemelli74:
Thx!

Will merge the day before beta cut for 2025.8

jamesonuk (approved):
Run a few tests clean and existing and all works.
Looks good.

library_test.py
```diff
@@ -155,21 +156,26 @@ async def main() -> None:
...
+        client_session,
         args.country,
         args.email,
         args.password,
         login_data_stored,
-        args.save_raw_data,
     )
 
+    if args.save_raw_data:
+        api.save_raw_data()
```
bdraco, line 170: Need to check if this is async-safe

bdraco: `_save_to_file` is not async safe and should be running in the executor. And it may need an asyncio.Lock around it if it can be called multiple times

chemelli74: This is a debug only call and is not used at all in HomeAssistant but only from library_test.py command line tools.

PR #388, opened 2025-07-19
fix: using default=True test cannot be disabled

ivanfmartinez:
The current option are enabled by default and cannot be disabled, the store_true action will put True again and cant be disabled.

This make easier for testing the login process only without testing the devices when not need.

### DIFF
--- a/library_test.py
+++ b/library_test.py
@@ -159,7 +159,10 @@ async def main() -> None:
     login_data_stored = read_from_file(args.login_data_file)
 
     if not login_data_stored and not args.password:
-        print("You have to specify credentials for ", args.email if sys.stdout.isatty() else obfuscate_email(args.email))
+        print(
+            "You have to specify credentials for ",
+            args.email if sys.stdout.isatty() else obfuscate_email(args.email),
+        )
         args.password = getpass.getpass("Password: ")
 
     api = AmazonEchoApi(
@@ -197,7 +200,9 @@ async def main() -> None:
     print("Logged-in.")
 
     print("-" * 20)
-    print("Login data:", login_data if sys.stdout.isatty() else scrub_fields(login_data))
+    print(
+        "Login data:", login_data if sys.stdout.isatty() else scrub_fields(login_data)
+    )
     print("-" * 20)
 
     save_to_file(f"{SAVE_PATH}/output-login-data.json", login_data)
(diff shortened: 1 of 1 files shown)


chemelli74:
I use `"test": false` in my library_test.json since long time and it just works fine.

Can you please show the command line or the json you are using ?

ivanfmartinez:
> I use `"test": false` in my library_test.json since long time and it just works fine.
> 
> Can you please show the command line or the json you are using ?

I'm trying to set it from command line and not on the file. If you consider that the correct method is only the json file I think the command line option should be removed, beucase it does not work now.

The current code with default=True for the commandline will not work because using --test with put True on a value that already has True, store_true should be used with a default of False.

the json file are allways ovewriting the command line when the entry is defined on json.

library_test.py
```diff
@@ -65,16 +66,20 @@ def get_arguments() -> tuple[ArgumentParser, Namespace]:
...
         help="Save HTML source on disk",
     )
     parser.add_argument(
         "--test",
         "-t",
         action="store_true",
-        default=True,
         help="Execute test actions",
     )
+    parser.add_argument(
```
chemelli74, line 77: This is a new feature, please remove from this fix PR

library_test.py
```diff
@@ -154,7 +159,10 @@ async def main() -> None:
     login_data_stored = read_from_file(args.login_data_file)
 
     if not login_data_stored and not args.password:
-        print(f"You have to specify credentials for {args.email}")
+        print(
+            "You have to specify credentials for ",
+            args.email if sys.stdout.isatty() else obfuscate_email(args.email),
```
chemelli74, line 164: Same as above, one topic per PR

library_test.py
```diff
@@ -192,11 +200,18 @@ async def main() -> None:
     print("Logged-in.")
 
     print("-" * 20)
-    print("Login data:", login_data)
+    print(
+        "Login data:", login_data if sys.stdout.isatty() else scrub_fields(login_data)
```
chemelli74, line 204: Here as well

ivanfmartinez:
replaced by https://github.com/chemelli74/aioamazondevices/pull/390

chemelli74:
> replaced by https://github.com/chemelli74/aioamazondevices/pull/390

No need to close a PR next time, just fix it ;-)

PR #389, opened 2025-07-19
feat: add Server-Sent-Events via HTTP2

chemelli74:
Initial code from @chemelli74 

Updated by @francescolf and @jamesonuk 

## Summary by CodeRabbit

* **New Features**
  * HTTP/2 push-event streaming with background processing and device capability registration
  * Matter device detection event

* **Improvements**
  * More robust multipart JSON parsing and ignore-empty-capabilities responses
  * Volume tracking refined (muted can be unknown) and duplicate notifications suppressed
  * Graceful Ctrl-C shutdown sequence, background task cleanup, expanded sensitive-data redaction, and reduced noisy logging levels


src/aioamazondevices/api.py
```diff
@@ -1361,3 +1423,135 @@ async def _get_dnd_status(self) -> dict[str, AmazonDeviceSensor]:
...
+            return
+
+        if not self._login_stored_data.get(DEVICE_CAPABILITIES_REGISTERED, False):
+            _LOGGER.warning("Device capabilities not set, registering now.")
+            await self._register_device_capabilities()
+
+        await self._http2_init_client()
+
+        try:
+            async with self._http2_client.stream(
```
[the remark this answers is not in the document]
bdraco: Something like
```python

      try:
          async with aconnect_sse(
              self._http2_client,
              "GET",
              f"{await self._http2_site()}/v{HTTP2_DIRECTIVES_VERSION}/directives",
              headers={
                  "Authorization": f"Bearer {self._login_stored_data['access_token']}",
              },
          ) as event_source:
              async for sse in event_source.aiter_sse():
                  _LOGGER.debug("SSE event: %s, data: %s", sse.event, sse.data)

                  # Handle ping events
                  if sse.event == "ping" or not sse.data:
                      await self._ping()
                      continue

                  try:
                      # Parse SSE data field as JSON
                      chunk_json = await self._parse_directive_data(sse.data)
                      # ... process chunk_json ...
                  except (orjson.JSONDecodeError, KeyError, IndexError) as err:
                      _LOGGER.warning("Failed to parse directive: %s", err)
                      continue

      except httpx.HTTPError as excp:
          _LOGGER.warning("SSE connection error: %s", excp
```

src/aioamazondevices/implementation/http2.py
```diff
@@ -0,0 +1,283 @@
...
+                    _LOGGER.debug(
+                        "AVS Directives response status: %s [%s]",
+                        response.status_code,
+                        response.http_version,
+                    )
+                    async for chunk in response.aiter_text():
+                        if self._stop_event.is_set():
+                            break
+                        _LOGGER.debug("AVS Directives chunk: %s", chunk)
+                        if chunk.startswith("-----"):
```
bdraco: manual SSE parsing is fragile and will likely break on edge cases (multi-part boundaries, partial chunks, varying content-type headers).

src/aioamazondevices/api.py
```diff
@@ -161,6 +168,20 @@ async def _get_sensors_states(self) -> dict[str, dict[str, AmazonDeviceSensor]]:
 
         return devices_sensors
 
+    async def start_http2_thread(self) -> None:
+        """Start HTTP2 background thread."""
+        await self._http2_client.start_thread()
+
```
bdraco: There is no stop_http2_thread so graceful cleanup never happens

pyproject.toml
```diff
@@ -24,6 +24,7 @@ dependencies = [
 "anyio",
```
bdraco: Doesn't look like this is every used directly

pyproject.toml
```diff
@@ -24,6 +24,7 @@ dependencies = [
...
 "beautifulsoup4",
 "colorlog",
+"httpx",
```
bdraco: "h2" is now a required dep

src/aioamazondevices/implementation/http2.py
```diff
@@ -0,0 +1,283 @@
...
+                        chunk_device = (
+                            updates_node["resourceMetadata"]["payload"]
+                            .get("dopplerId", {})
+                            .get("deviceSerialNumber")
+                        )
+                        payload = updates_node.get("resourceMetadata", {}).get(
+                            "payload", {}
+                        )
+
+                        if chunk_type in AmazonPushMessage.__members__.values():
```
bdraco: could be in `AmazonPushMessage._value2member_map_`

bdraco:
Top item is the fragile parser.  Its almost guaranteed to have unexpected bugs in production. A custom parser should live in its own module and have tests. Even a vibe coded one would likely be better than the manual parsing one. The SSE format is well known and a tool can likely put one together that better handles edge cases
https://github.com/chemelli74/aioamazondevices/pull/389/changes#r2880966700

The leaks and lack of stop are the other big stand outs.

Everything else is mostly nits

pyproject.toml
chemelli74: This is added by poetry in poetry.lock

src/aioamazondevices/api.py
```diff
@@ -170,6 +181,30 @@ async def get_devices_data(
...
+            serial = payload.get("dopplerId", {}).get("deviceSerialNumber")
+            if serial:
+                self._volume_states[serial] = AmazonVolumeState(
+                    payload.get("volumeSetting"), bool(payload.get("isMuted"))
+                )
+            await self._emit_volume_state_event()
+            return
+        if event_type == AmazonPushMessage.AudioPlayerState.value:
+            await self.sync_media_state()
+            return
```
chemelli74, line 246: do we miss _emit_media_state_event ?

IF the method is used once, as it's only 2 lines of code, we may want to collapse it here

src/aioamazondevices/implementation/http2.py
```diff
@@ -0,0 +1,281 @@
...
+        session_state_data: AmazonSessionStateData,
+    ) -> None:
+        """Initialize Amazon HTTP2 client class."""
+        self._http_wrapper = http_wrapper
+        self._login_stored_data = session_state_data.login_stored_data
+
+        self._http2_client: httpx.AsyncClient
+        self.on_push_event: Signal[str, dict[str, Any]] = Signal(self)
+
+        self.reconnect_delay = HTTP2_RECONNECT_DELAY
```
chemelli74:
```suggestion
        self._reconnect_delay = HTTP2_RECONNECT_DELAY
```

or even drop and use HTTP2_RECONNECT_DELAY directly

src/aioamazondevices/implementation/http2.py
```diff
@@ -0,0 +1,281 @@
...
+            with contextlib.suppress(asyncio.CancelledError):
+                await self._task
+
+        if self._pending_push_tasks:
+            for t in list(self._pending_push_tasks):
+                t.cancel()
+            await asyncio.gather(*self._pending_push_tasks, return_exceptions=True)
+            self._pending_push_tasks.clear()
+
+    def is_connected(self) -> bool:
```
chemelli74: property ?

src/aioamazondevices/implementation/http2.py
```diff
@@ -0,0 +1,281 @@
...
+        _LOGGER.debug("Initialized HTTP2 client")
+
+    def _http2_site(self) -> str:
+        """Get HTTP2 site."""
+        if not self._login_stored_data:
+            _LOGGER.debug("No login data available, cannot get HTTP2 site")
+            return ""
+
+        region = self._login_stored_data["customer_info"]["home_region"]
+        return HTTP2_SITE.replace("{region}", region)
```
chemelli74:
```suggestion
        return HTTP2_SITE.format(region=region)
```

Probably cleaner

src/aioamazondevices/implementation/http2.py
```diff
@@ -0,0 +1,281 @@
...
+            HTTPStatus.FORBIDDEN,
+            HTTPStatus.PROXY_AUTHENTICATION_REQUIRED,
+            HTTPStatus.UNAUTHORIZED,
+        ):
+            raise CannotAuthenticate(
+                "Detected ping 403, please check your credentials and region"
+            )
+
+    async def _build_ssl_context_async(self) -> SSLContext:
+        return await asyncio.to_thread(create_default_context)
```
chemelli74: One line method used only once, probably not worth it

src/aioamazondevices/api.py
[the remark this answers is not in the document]
jamesonuk: These are not SSE events.    These is actually a stream of `multipart/related` MIME messages

src/aioamazondevices/implementation/http2.py
```diff
@@ -0,0 +1,450 @@
...
+    async def stop_processing(self) -> None:
+        """Stop all background tasks gracefully."""
+        self._stop_event.set()
+        self._connected_event.clear()
+
+        for task in (self._stream_task, self._ping_task):
+            if task and not task.done():
+                task.cancel()
+                with contextlib.suppress(asyncio.CancelledError):
+                    await task
```
bdraco: this will swallow cancellation

bdraco:
```suggestion
try:
   await task
except asyncio.CancelledError:
  if (current_task := asyncio.current_task()) and current_task.cancelling():
     raise
```

needs formatting

src/aioamazondevices/implementation/http2.py
```diff
@@ -0,0 +1,450 @@
...
+            ):
+                await asyncio.sleep(HTTP2_RECONNECT_DELAY)
+                continue
+
+            try:
+                await self._stream_and_process()
+            except httpx.RemoteProtocolError as exc:
+                _LOGGER.debug("HTTP2 disconnection detected: %s", exc)
+            except httpx.HTTPError as exc:
+                _LOGGER.warning("HTTP2 error detected: %s", exc)
```
bdraco, line 256: broad except + log + raise makes sense so unexpected exceptions don't silently kill the task

src/aioamazondevices/implementation/http2.py
```diff
@@ -0,0 +1,450 @@
...
+
+    async def _get_avs_directives(self) -> None:
+        """Maintain AVS directive stream loop."""
+        while not self._stop_event.is_set():
+            self._connected_event.clear()
+
+            if not (
+                await self._refresh_token()
+                and await self._check_device_capabilities_registered()
+            ):
```
bdraco: if their of these raise unexpected exceptions the task will collapse without logging, catch Exception, LOGGER.execption, and re-raise

src/aioamazondevices/implementation/http2.py
```diff
@@ -0,0 +1,450 @@
...
+        self._stop_event.clear()
+        self._connected_event.clear()
+        self._stream_task = asyncio.create_task(
+            self._get_avs_directives(), name="avs-stream"
+        )
+        self._ping_task = asyncio.create_task(self._ping_loop(), name="avs-ping")
+
+    async def stop_processing(self) -> None:
+        """Stop all background tasks gracefully."""
+        self._stop_event.set()
```
bdraco: doesn't reset the client so restarting is not possible. not sure if its a problem but its a code smell

src/aioamazondevices/implementation/http2.py
```diff
@@ -0,0 +1,450 @@
...
+            return
+
+        rendering_updates = AmazonHTTP2Client._extract_rendering_updates(chunk_json)
+        if rendering_updates is None:
+            return
+
+        # All observed messages only contain one update but it is presented
+        # as an array so iterate over all results in case that ever changes.
+        for rendering_update in rendering_updates:
+            result = AmazonHTTP2Client._process_rendering_update(rendering_update)
```
bdraco: calling protected methods here is a bit of code smell

src/aioamazondevices/implementation/http2.py
```diff
@@ -0,0 +1,450 @@
...
+                self._connected_event.clear()
+
+    async def _handle_part(self, part: bytes) -> None:
+        """Process a single multipart section."""
+        if not part:
+            with contextlib.suppress(Exception):
+                await self._ping()
+            return
+
+        chunk_json = AmazonHTTP2Client._extract_json_from_part(part)
```
bdraco: This can raise ValueError and raise upward

src/aioamazondevices/implementation/http2.py
```diff
@@ -0,0 +1,450 @@
...
+
+    def _get_bearer_token(self) -> str:
+        """Return the current access token."""
+        if not (
+            token := self._session_state_data.login_stored_data.get(
+                REFRESH_ACCESS_TOKEN
+            )
+        ):
+            _LOGGER.error("No access token available")
+            raise CannotAuthenticate("No access token available")
```
bdraco, line 451: raises upward but not all callers handle it

src/aioamazondevices/implementation/http2.py
```diff
@@ -0,0 +1,450 @@
...
+            }
+
+        if isinstance(obj, list):
+            return [AmazonHTTP2Client._string_recursive_parse(i) for i in obj]
+
+        if isinstance(obj, str) and obj.startswith(("{", "[")):
+            try:
+                return AmazonHTTP2Client._string_recursive_parse(orjson.loads(obj))
+            except orjson.JSONDecodeError:
+                return obj
```
bdraco: be sure you trust your input here as its a vuln if ever run on untrusted input because it could recurse an use up all ram. work a comment in the docstring that you need to trust the source data

src/aioamazondevices/implementation/http2.py
```diff
@@ -0,0 +1,450 @@
...
+            finally:
+                _LOGGER.debug("AVS Directives stream closed")
+                self._connected_event.clear()
+
+    async def _handle_part(self, part: bytes) -> None:
+        """Process a single multipart section."""
+        if not part:
+            with contextlib.suppress(Exception):
+                await self._ping()
+            return
```
bdraco: I'm not sure whats going on here but it looks like there is already a ping loop. Why is this here as well?  Also its swallowing exceptions which will lead to hard to find bugs

src/aioamazondevices/implementation/http2.py
```diff
@@ -0,0 +1,450 @@
...
+            if task and not task.done():
+                task.cancel()
+                with contextlib.suppress(asyncio.CancelledError):
+                    await task
+
+    async def _register_device_capabilities(self) -> None:
+        """Register device capabilities."""
+        _, raw_resp = await self._http_wrapper.session_request(
+            method=HTTPMethod.PUT,
+            url=f"https://api.amazonalexa.com{URI_CAPABILITIES}",
```
bdraco, line 223: no region?

jamesonuk: > no region?

Yep that is correct.   This is a global call as seen via the app.

src/aioamazondevices/implementation/http2.py
```diff
@@ -0,0 +1,450 @@
...
+                response.http_version,
+            )
+
+            if response.status_code in (
+                HTTPStatus.UNAUTHORIZED,
+                HTTPStatus.FORBIDDEN,
+            ):
+                raise CannotAuthenticate(
+                    f"Directives stream returned {response.status_code}"
+                )
```
bdraco, line 320: probably should clear `login_stored_data[DEVICE_CAPABILITIES_REGISTERED] = False` when it gets an auth failure or it likely can't ever recover.. ok if a new client gets created when creds fail though

src/aioamazondevices/api.py
```diff
@@ -178,6 +184,47 @@ async def get_devices_data(
...
+        """Stop HTTP2 background thread."""
+        if self._http2_client:
+            await self._http2_client.stop_processing()
+
+    async def _http2_push_event_handler(
+        self, event_type: str, payload: dict[str, Any]
+    ) -> None:
+        _LOGGER.debug("Event - %s : Payload - %s", event_type, payload)
+        if event_type == AmazonPushMessage.VolumeChange.value:
+            serial = payload.get("dopplerId", {}).get("deviceSerialNumber")
```
bdraco, line 229: nit: assumes the payload is well formed here and will raise if `dopplerId` is not a `dict`, but than you probably have bigger problems if amazon changes the payload type.. maybe need to log better on failure

src/aioamazondevices/implementation/http2.py
jamesonuk: > be sure you trust your input here as its a vuln if ever run on untrusted input because it could recurse an use up all ram. work a comment in the docstring that you need to trust the source data

https://github.com/chemelli74/aioamazondevices/blob/8e24732497c379ec7bf4c6f71c161ed6b5d6fd21/src/aioamazondevices/implementation/http2.py#L203

should restrict the maximum size to 512KB and internal method will only be used to parse Amazon responses.

src/aioamazondevices/implementation/http2.py
jamesonuk: They are `@staticmethod` and internal.   Is there a better naming convention for these?

bdraco: drop the underscore to indicate its ok to call them from outside itself or subclasses

src/aioamazondevices/implementation/http2.py
jamesonuk: The client is passed in to library in https://github.com/chemelli74/aioamazondevices/blob/8e24732497c379ec7bf4c6f71c161ed6b5d6fd21/src/aioamazondevices/api.py#L187 so the calling client is responsible for clearing up.
When restarting a client reference will be passed but rather than doing anything here we should probably set
`self.__http2_client = None` when stopping here
https://github.com/chemelli74/aioamazondevices/blob/8e24732497c379ec7bf4c6f71c161ed6b5d6fd21/src/aioamazondevices/api.py#L208-L209 ??

src/aioamazondevices/implementation/http2.py
jamesonuk: > I'm not sure whats going on here but it looks like there is already a ping loop. Why is this here as well? Also its swallowing exceptions which will lead to hard to find bugs

@chemelli74 this is the ping that takes place upon receipt of the first part (with no content).   This is outside of the standard ping loop.   Was this something that we needed to respond to the initial request with a ping or can this now be removed completely as we have the ping loop?

src/aioamazondevices/implementation/http2.py
jamesonuk: > broad except + log + raise makes sense so unexpected exceptions don't silently kill the task

as in `exception Exception:` ?

bdraco:
```python
except Exception:
   LOGGER.exception("useful message here.. about the unexpected exception while doing X")
   raise # <-- be sure to re-raise for unknown exception
```

Right now it relies on callers propagating it or having an unhandled exception hook which would give a worse message

src/aioamazondevices/implementation/http2.py
jamesonuk: The capability registration is a one off task (that essentially registers the serial number server side as capable of receiving messages).    This would only change if you created a new instance of the integration (in which case a new registration would be made)

src/aioamazondevices/api.py
jamesonuk: Yeah I think there are bigger issues if that happens.   If it does then it should be noticed pretty quickly and enabling debug logging would trigger the previous line.

src/aioamazondevices/implementation/http2.py
jamesonuk: now without outer exception block

src/aioamazondevices/implementation/http2.py
jamesonuk: implemented

jamesonuk: implemented changed as below

src/aioamazondevices/implementation/http2.py
jamesonuk: all `ValueErrror`s are not handled in `_extract_json_from_part`

src/aioamazondevices/implementation/http2.py
jamesonuk: Now logged with generic `Exception` handler and then re-raised

src/aioamazondevices/implementation/http2.py
chemelli74: We can remove it completely now that there is a proper ping loop

jamesonuk:
@coderabbitai review

src/aioamazondevices/implementation/http2.py
```diff
@@ -0,0 +1,501 @@
...
+            parsed = orjson.loads(body)
+            return cast(
+                "dict[str, Any]", AmazonHTTP2Client._string_recursive_parse(parsed)
+            )
+        except (TypeError, ValueError, orjson.JSONDecodeError) as exc:
+            _LOGGER.warning(
+                "Failed to parse multipart section: %s",
+                part.decode("utf-8", errors="replace"),
+                exc_info=exc,
+            )
```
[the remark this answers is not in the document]
jamesonuk: @chemelli74 do we want to push this to debug or save that until it is widely tested?
Just wary that if this does every come up that I suspect it may be difficult to reproduce and if debug logging was off it would be gone.    That said what are the changes of this actually occurring?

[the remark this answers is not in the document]
chemelli74: @jamesonuk I think we should really log this as it may help a lot in case of failure. We will switch to debug only once it's mature and very well tested.

src/aioamazondevices/implementation/http2.py
jamesonuk: No longer as issue as http client is provided by caller now

src/aioamazondevices/api.py
jamesonuk: as discussed, this matches volume pattern and is likely to be used elsewhere when we move on to looking at the next stage of reviewing how the http2 event data is used.   Can revist then

src/aioamazondevices/implementation/http2.py
jamesonuk: In addition I have added docstring about only using for trusted sources and introduced a max depth of 10 which combined with the 512kb limit should all but stop any resource exhaustion issues

src/aioamazondevices/implementation/http2.py
```diff
@@ -0,0 +1,422 @@
...
+    async def stop_processing(self) -> None:
+        """Stop all background tasks gracefully."""
+        async with self._process_lock:
+            self._stop_event.set()
+            self._connected_event.clear()
+            if self._run_task and not self._run_task.done():
+                self._run_task.cancel()
+                with contextlib.suppress(asyncio.CancelledError, Exception):
+                    await self._run_task
+
```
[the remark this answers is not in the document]
jamesonuk: @coderabbitai this is within `stop_processing` and after we have attempted to cancel the task.

I do not see any point in propagating exceptions at this point as the decision has already been made to stop the task

chemelli74:
@coderabbitai review

src/aioamazondevices/implementation/http2.py
```diff
@@ -0,0 +1,438 @@
...
+    async def _refresh_token(self) -> bool:
+        """Refresh the access token."""
+        try:
+            refreshed_token, _ = await self._http_wrapper.refresh_data(
+                REFRESH_ACCESS_TOKEN
+            )
+            if not refreshed_token:
+                _LOGGER.warning("Failed to refresh access token")
+                return False
+        except (CannotConnect, CannotRetrieveData) as exc:
```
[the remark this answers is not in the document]
jamesonuk: This is correct.   If authentication fails a retry will not work.   A reference to the task has been returned to the caller and it is up to the caller to use the `on_done_callback` to handle exceptions

src/aioamazondevices/api.py
```diff
@@ -178,6 +185,53 @@ async def get_devices_data(
...
+            if serial:
+                volume = AmazonVolumeState(
+                    payload.get("volumeSetting"), payload.get("isMuted")
+                )
+                self._media_handler.update_cached_device_volume(serial, volume)
+            await self._emit_volume_state_event()
```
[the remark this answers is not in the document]
jamesonuk: Not an issue in practice as we call sync_media before starting http2 loop in HA.
https://github.com/home-assistant/core/pull/165825/changes#diff-44a9e899fbe6884b43218464aa388afb14ca2bfa1e380e29ccf5158ded56bb3bR43
From a library point of view it may still be worth adding this check?

src/aioamazondevices/implementation/http2.py
```diff
@@ -0,0 +1,438 @@
...
+                    and await self._check_device_capabilities_registered()
+                ):
+                    await asyncio.sleep(HTTP2_RECONNECT_DELAY)
+                    continue
+
+                await self._stream_and_process()
+            except httpx.RemoteProtocolError as exc:
+                _LOGGER.debug("HTTP2 disconnection detected: %s", exc)
+            except httpx.HTTPError as exc:
+                _LOGGER.warning("HTTP2 error detected: %s", exc)
```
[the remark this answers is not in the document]
jamesonuk: As above, task reference is returned to caller and it is up to them to handle authentication issues.   Simple retry won't work.

src/aioamazondevices/api.py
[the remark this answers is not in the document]
chemelli74: I think is a small change and makes logic more clear.

chemelli74:
@coderabbitai review

src/aioamazondevices/http_wrapper.py
```diff
@@ -373,8 +377,9 @@ async def session_request(
...
             ]:
                 raise CannotAuthenticate(await self.http_phrase_error(resp.status))
-            if not await self._ignore_ap_signin_error(resp):
-                _LOGGER.debug("Error response content: %s", await resp.text())
+            if not await self._ignore_ap_signin_error(
+                resp
+            ) and not await self._ignore_capabilities_error(resp):
                 raise CannotRetrieveData(
                     f"Request failed: {await self.http_phrase_error(resp.status)}"
                 )
```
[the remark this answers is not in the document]
chemelli74: @copilot , we raise the CannotRetrieveData exception only for two endpoints that we know return non-200 responses (ap_signal and capabilities); for all others, nothing has changed.

library_test.py
```diff
@@ -325,6 +311,22 @@ def _on_http2_task_done(task: asyncio.Task[None]) -> None:
...
+        http2_task = await api.start_http2_processing(httpx_client)
+
+        def _on_http2_task_done(task: asyncio.Task[None]) -> None:
+            if not task.cancelled() and task.exception():
+                print("HTTP2 processing task ended with exception:")
+                print(task.exception())
+
+        http2_task.add_done_callback(_on_http2_task_done)
+
+        await wait_until_ctrl_c()
```
chemelli74: There are 2 calls to this method now

jamesonuk: This was to give separation to what is going on.   I was finding it hard to read the logs as http2 setup calls were overlapping with that device setup calls.

No we have all the original device stuff.   CTRL-C then starts http2 processing and you can keep going as long as you want to see http2 events and then CTRL-C a second time to start the tests.

chemelli74: The start testing should not be linked to a ctrl+c ;-)

You can always mute/unmute them based on json settings if you prefer to have a focus on a specific test group

jamesonuk: > The start testing should not be linked to a ctrl+c ;-)
> 
> You can always mute/unmute them based on json settings if you prefer to have a focus on a specific test group

Testing already was
https://github.com/chemelli74/aioamazondevices/blob/7b70b61a52a7e9daa60f6e07bd41d7353cae5fef/library_test.py#L325-331

The issue is that you were seeing all the http2 startup logging intertwined with the get devices data and rest setup calls.   You would then see volume / EQ... Events whilst waiting for ctrl+C

Just made it difficult to verify the device data (which I guess was the point of adding the Ctrl+C break) and the starting of the http2 loops as due to the length of the some of the payloads it quickly drops off the buffer.

chemelli74: Ctrl-c was there to exit http2 infinite loop.
Only for that, no other reason

jamesonuk (approved):

src/aioamazondevices/const/http.py
```diff
@@ -43,3 +44,7 @@
 URI_NEXUS_GRAPHQL = "/nexus/v1/graphql"
 URI_NOTIFICATIONS = "/api/notifications"
 URI_SIGNIN = "/ap/signin"
+
+HTTP2_SITE = "https://alexa.{region}.gateway.devices.a2z.com"
```
francescolf, line 48: We might want to stick to what the app is using and change this to `https://bob-dispatch-prod-{region}.amazon.com`

chemelli74: iOS use the coded one. LEt's monitor in future and see if we need to change it

PR #393, opened 2025-07-21
fix: change country logic for fixing Brasil and Netherland

chemelli74:
- Fixes #392

PR #400, opened 2025-07-22
fix change country check from marketPlaceDomainName to CountryOfResidence

jamesonuk:
use `countryOfResidence` rather than `marketPlaceDomainName`


chemelli74:
Let's park this for a while.
Seems we sorted out all remaining countries but Australia with #393

chemelli74:
Superseded by #449

PR #412, opened 2025-07-27
feat: add volume control

jamesonuk:


src/aioamazondevices/api.py
```diff
@@ -1134,3 +1140,19 @@ async def set_do_not_disturb(self, device: AmazonDevice, state: bool) -> None:
...
+        """Call Alexa.DeviceControls.Volume to set volume."""
+        if volume < MIN_VOLUME:
+            _vol_clean = MIN_VOLUME
+        if volume > MAX_VOLUME:
+            _vol_clean = MAX_VOLUME
+        else:
+            _vol_clean = math.ceil(volume / 10.0) * 10
+        if volume != _vol_clean:
+            _LOGGER.debug("Volume %s rounded to %s", volume, _vol_clean)
+        return await self._send_message(device, AmazonSequenceType.Volume, _vol_clean)
```
chemelli74, line 1158:
```suggestion
        return await self._send_message(device, AmazonSequenceType.Volume, str(_api_volume))
```

and avoid changing the method typing

src/aioamazondevices/api.py
```diff
@@ -940,7 +944,7 @@ async def _send_message(
         self,
         device: AmazonDevice,
         message_type: str,
-        message_body: str,
+        message_body: str | float,
```
chemelli74, line 947:
```suggestion
        message_body: str,
```

I would avoid changing the method typing

src/aioamazondevices/api.py
```diff
@@ -1030,7 +1036,7 @@ async def _send_message(
                     "deviceSerialNumber": device.serial_number,
                 },
                 "connectionRequest": {
-                    "uri": "connection://AMAZON.Launch/" + message_body,
+                    "uri": "connection://AMAZON.Launch/" + str(message_body),
```
chemelli74, line 1039:
```suggestion
                    "uri": "connection://AMAZON.Launch/" + message_body,
```

src/aioamazondevices/api.py
```diff
@@ -1134,3 +1140,19 @@ async def set_do_not_disturb(self, device: AmazonDevice, state: bool) -> None:
...
         )
+
+    async def set_volume(
+        self,
+        device: AmazonDevice,
+        volume: int,
+    ) -> None:
+        """Call Alexa.DeviceControls.Volume to set volume."""
+        if volume < MIN_VOLUME:
+            _vol_clean = MIN_VOLUME
```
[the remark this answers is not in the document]
jamesonuk: but it needs to be a public method rather than internal?

chemelli74 (changes requested):

PR #415, opened 2025-07-27
fix: batch sequence operations

jamesonuk:
this opens up a number of opportunities to combine operations that will reduce number of calls (less 429s) but also perform more complex logic.

Eg. a good use case would be to accept a volume argument for the notify action and have the one `send_message`
* Set the volume to desired level
* Send announcement
* Reset the volume to previous level

Another good use would be if someone is using the HA notify action to send the same announcement to multiple devices we should convert this into one message which would cut the 429s and improve response

## Summary by CodeRabbit

* **New Features**
  * Message batching for single devices and simultaneous sending to device clusters.
  * Optimised grouping of similar operations to reduce duplicate work.

* **Refactor**
  * Message sending now uses asynchronous batching with a short delay (≈0.3s) to improve throughput and reliability.


src/aioamazondevices/api.py
```diff
@@ -1181,51 +1211,46 @@ async def _send_message(
...
+    async def _batch_sequence_calls(self) -> None:
+        await asyncio.sleep(0.3)
+        async with self._sequence_lock:
+            nodes = self._sequence_buffer
+            self._sequence_buffer = []
+            self._sequence_build_running = False
+
+        if nodes:
+            _LOGGER.debug("Batching %i sequences", len(nodes))
+            await self._send_sequences(nodes)
```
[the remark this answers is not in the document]
bdraco: This is actually a decent point

src/aioamazondevices/api.py
```diff
@@ -1181,51 +1211,46 @@ async def _send_message(
...
+                }
+            )
+            if not self._sequence_build_running:
+                self._sequence_build_running = True
+                task = asyncio.create_task(self._batch_sequence_calls())
+                self._sequence_tasks.add(task)
+                task.add_done_callback(self._sequence_tasks.discard)
+
+    async def _batch_sequence_calls(self) -> None:
+        await asyncio.sleep(0.3)
```
bdraco: Hard coded magic number

src/aioamazondevices/api.py
```diff
@@ -1181,51 +1211,46 @@ async def _send_message(
...
+                    "@type": "com.amazon.alexa.behaviors.model.OpaquePayloadOperationNode",  # noqa: E501
+                    "type": message_type,
+                    "operationPayload": payload,
+                }
+            )
+            if not self._sequence_build_running:
+                self._sequence_build_running = True
+                task = asyncio.create_task(self._batch_sequence_calls())
+                self._sequence_tasks.add(task)
+                task.add_done_callback(self._sequence_tasks.discard)
```
bdraco: Having all of this inside this object makes the flow hard to follow.

I'd make a new `SequenceRunner` object and abstract all the logic behind that so its in charge of the background task and synchronization. You should only need one at a time instead of a set of them if your locking is correct

jamesonuk: @bdraco I have not used runners before.   Can you point me towards somewhere that implements this or give me a quick bit of pseudocode for what you are suggesting

bdraco:
Had a chance to look at this again and a few things still stand out:

- ~~No way to shutdown cleanly or know when the batch has been sent. Maybe an `await obj.flush()`?~~
- ~~Errors are still silently swallowed, should at least log since the caller no longer gets feedback~~

bdraco:
> * Errors are still silently swallowed, should at least log since the caller no longer gets feedback

O I see you answered at https://github.com/chemelli74/aioamazondevices/pull/415#discussion_r2455447238 and there is a LOGGER.exception so I guess thats the best we can do

bdraco:
> Had a chance to look at this again and a few things still stand out:
> 
> * No way to shutdown cleanly or know when the batch has been sent. Maybe an `await obj.flush()`?

But that probably doesn't matter since if you are shutting down its ok for it to just get GCed away

bdraco:
Is the nesting correct.

[demo_routines_bug.py](https://github.com/user-attachments/files/26663157/demo_routines_bug.py)
```bash
% uv run python demo_routines_bug.py 
======================================================================
1) Normal speak — correct structure
======================================================================
{
  "@type": "com.amazon.alexa.behaviors.model.Sequence",
  "startNode": {
    "@type": "com.amazon.alexa.behaviors.model.SerialNode",
    "nodesToExecute": [
      {
        "@type": "com.amazon.alexa.behaviors.model.OpaquePayloadOperationNode",
        "type": "Alexa.Speak",
        "operationPayload": {
          "textToSpeak": "Hello"
        }
      }
    ]
  }
}

======================================================================
2) Routine via new code — SerialNode nested inside SerialNode
======================================================================
{
  "@type": "com.amazon.alexa.behaviors.model.Sequence",
  "startNode": {
    "@type": "com.amazon.alexa.behaviors.model.SerialNode",
    "nodesToExecute": [
      {
        "@type": "com.amazon.alexa.behaviors.model.SerialNode",
        "nodesToExecute": [
          {
            "@type": "com.amazon.alexa.behaviors.model.OpaquePayloadOperationNode",
            "type": "Alexa.Speak",
            "operationPayload": {
              "textToSpeak": "Good morning"
            }
          },
          {
            "@type": "com.amazon.alexa.behaviors.model.OpaquePayloadOperationNode",
            "type": "Alexa.Sound",
            "operationPayload": {
              "soundStringId": "bell_02"
            }
          }
        ]
      }
    ]
  }
}

======================================================================
3) What the OLD code sent for routines — the sequence as-is
======================================================================
{
  "@type": "com.amazon.alexa.behaviors.model.Sequence",
  "startNode": {
    "@type": "com.amazon.alexa.behaviors.model.SerialNode",
    "nodesToExecute": [
      {
        "@type": "com.amazon.alexa.behaviors.model.OpaquePayloadOperationNode",
        "type": "Alexa.Speak",
        "operationPayload": {
          "textToSpeak": "Good morning"
        }
      },
      {
        "@type": "com.amazon.alexa.behaviors.model.OpaquePayloadOperationNode",
        "type": "Alexa.Sound",
        "operationPayload": {
          "soundStringId": "bell_02"
        }
      }
    ]
  }
}
```

src/aioamazondevices/implementation/sequence.py
```diff
@@ -28,23 +44,92 @@ def __init__(
...
+            sequences: List of sequence nodes to optimise
+
+        """
+        # list will be in the order nodes were added
+        # group by message type, body and music provider id to find similar operations
+        # then wrap in parallel node if for different devices
+        for _, group_iter in groupby(
+            sequences,
+            key=lambda x: (x.message_type, x.message_body, x.music_provider_id),
+        ):
```
bdraco, line 108: This look like it won't work correctly if someone does 

```python                            
  await api.call_alexa_speak(dev1, "hello")                                                                                                                                                                                                        
  await api.call_alexa_speak(dev1, "goodbye")                                                                                                                                                                                                      
  await api.call_alexa_speak(dev2, "hello")
```

bdraco: I think you want something like

```python
  def _optimise_sequence_nodes(                                                                                                                                                                                                                    
      self, sequences: list[AmazonSequenceNode]
  ) -> Generator[dict[str, Any], None, None]:                                                                                                                                                                                                      
      groups: defaultdict[tuple, list[AmazonSequenceNode]] = defaultdict(list)
      for seq in sequences:                                                                                                                                                                                                                        
          groups[(seq.message_type, seq.message_body, seq.music_provider_id)].append(seq)
                                                                                                                                                                                                                                                   
      for group in groups.values():                                                                                                                                                                                                                
          if len(group) > 1 and len(
              {sequence.device.serial_number for sequence in group}                                                                                                                                                                                
          ) == len(group):                                                                                                                                                                                                                         
              yield {
                  "@type": "com.amazon.alexa.behaviors.model.ParallelNode",                                                                                                                                                                        
                  "nodesToExecute": [sequence.operation_node for sequence in group],
              }                                                                                                                                                                                                                                    
          else:
              yield from (sequence.operation_node for sequence in group)    
```

bdraco: if you need to support `speak("hello", dev1), speak("hello", dev2), speak("hello", dev1)`


```python
  def _optimise_sequence_nodes(                                                                                                                                                                                                                    
      self, sequences: list[AmazonSequenceNode]                                                                                                                                                                                                    
  ) -> Generator[dict[str, Any], None, None]:                                                                                                                                                                                                      
      groups: defaultdict[tuple, list[AmazonSequenceNode]] = defaultdict(list)                                                                                                                                                                     
      for seq in sequences:
          groups[(seq.message_type, seq.message_body, seq.music_provider_id)].append(seq)                                                                                                                                                          
                                                                                                                                                                                                                                                   
      for group in groups.values():
          # Split into rounds — each round has at most one item per device                                                                                                                                                                         
          remaining = list(group)
          while remaining:                                                                                                                                                                                                                         
              round_nodes: list[AmazonSequenceNode] = []
              seen_devices: set[str] = set()                                                                                                                                                                                                       
              leftover: list[AmazonSequenceNode] = []                                                                                                                                                                                              
              for node in remaining:
                  if node.device.serial_number not in seen_devices:                                                                                                                                                                                
                      seen_devices.add(node.device.serial_number)
                      round_nodes.append(node)                                                                                                                                                                                                     
                  else:
                      leftover.append(node)                                                                                                                                                                                                        
                            
              if len(round_nodes) > 1:
                  yield {
                      "@type": "com.amazon.alexa.behaviors.model.ParallelNode",                                                                                                                                                                    
                      "nodesToExecute": [n.operation_node for n in round_nodes],
                  }                                                                                                                                                                                                                                
              else:         
                  yield round_nodes[0].operation_node                                                                                                                                                                                              
   
              remaining = leftover   
```

jamesonuk:
> Is the nesting correct.

OK have been through, done some testing and checking and this works and is probably desired behaviour.
All sequences use routines but for most things it is a single step that is sent by HA.    With routines this is sending the whole routine.   You only get the nested `SerialNode` where the routine has multiple actions.

```
    "@type": "com.amazon.alexa.behaviors.model.SerialNode",
    "nodesToExecute": [
      {
        "@type": "com.amazon.alexa.behaviors.model.SerialNode",
        "nodesToExecute": [
```
works fine and actually keeps the routine in one atomic block.   Whilst unlikely it is possible that say an announcement could come along at the same time and be batched in the same HTTP call.   This would add the announcement to the outer `nodesToExecute`.

src/aioamazondevices/implementation/sequence.py
jamesonuk: I don't think this is something we need to worry about.

```
  await api.call_alexa_speak(dev1, "hello")
  await api.call_alexa_speak(dev1, "goodbye")
  await api.call_alexa_speak(dev2, "hello")
 ```
 could only happen within a poll period (0.3 seconds) so is pretty unlikely and if they arrive in this order then it would almost certainly have been separate HA calls.   In this case they would be queued and run sequentially.
 
 The changes are there to improve a few things:
 - message loss - if you send a long running speak command and then send a second whilst that is still playing the second can be swallowed by with no feedback (HTTP 200 response).
 - rate limiting - if you request lots of sequence calls in quick succession you get 429 responses (which are handled to some extend by an exponential backoff)
 - (re)applying the intent of one HA call with multiple targets.   If you send an announcement to three devices in one action call with three targets then I think we can assume these were intended to output simultaneously.   By the time these get to the integration we only see three separate calls and due to changes above they would all be queued sequentially.   The grouping is therefore to attempt to try and group the same call to multiple devices to run in parallel.
 
 These are ultimately working around the way Alexa works but not there to handle every scenario.   Given the batching is on such a short timeframe (and the order of the messages) I think this will be handled correctly

bdraco (approved):

chemelli74:
Thx @bdraco for the review and @jamesonuk for the work itself !!!

PR #444, opened 2025-08-13
fix: add proper configuration for Poland

maxmati:
I have tested it with my polish account.

### DIFF
--- a/src/aioamazondevices/const.py
+++ b/src/aioamazondevices/const.py
@@ -71,6 +71,7 @@ DOMAIN_BY_ISO3166_COUNTRY: dict[str, dict[str, Any]] = {
         "domain": "com.au",
         "openid.assoc_handle": f"{DEFAULT_ASSOC_HANDLE}_au",
     },
+    "pl": AMAZON_US_OVERRIDE,
     "tr": {
         "domain": "com.tr",
     },


jamesonuk:
@maxmati this change will logon to amazon.com but amazon.pl is a real site ???

in a browser where you are logged into amazon can you go to https://alexa.amazon.pl/api/users/me and confirm what it says for `countryOfResidence` and `effectiveMarketPlaceId` (top) and `marketPlaceDomainName` (bottom) or whether this page returns an error.

maxmati:
@jamesonuk yes it's is a thing but not fully. `https://alexa.amazon.pl/api/users/me` gives me "This site can’t be reached" "DNS_PROBE_FINISHED_NXDOMAIN" but when i go to `https://alexa.amazon.com/api/users/me` I get `"countryOfResidence":"PL"`, `"effectiveMarketPlaceId":"ATVPDKIKX0DER"` and `"marketPlaceDomainName":"https://www.amazon.com"`

maxmati:
# Market overwrite
I had to overwrite market because [default](https://github.com/chemelli74/aioamazondevices/blob/main/src/aioamazondevices/api.py#L136) is `["https://www.amazon.com"]` (notice an array) while correct for me was just `"https://www.amazon.com"`. Although this is topic for later since, as I understand, we have to figure out account configuration first.

# My account configuration

I've tried to follow this [this](https://github.com/chemelli74/aioamazondevices/wiki/Known-Issues-and-Limitations#country-details) guide but I had a few issues:

1.  I'm unable to access https://alexa.amazon.pl/api/users/me because domain alexa.amazon.pl doesn't exists ([dig result](https://digwebinterface.com/?hostnames=alexa.amazon.pl%0D%0A&type=&ns=resolver&useresolver=9.9.9.10&nameservers=))
2. When I open `https://www.amazon.pl/hz/mycd/preferences/myx#/home/settings/payment`  I get the message "Treści i preferencje urządzenia są konfigurowane za pomocą Amazon.com" (In rough translation "Content and devices preferences are configured through Amazon.com"
3. When I go to `https://www.amazon.com/hz/mycd/preferences/myx#/home/settings/payment` Then I can see "Current country/region Poland". I've tried to change it to Germany and then back to Poland but both changes had no effect on `marketPlaceDomainName`. I've been checking on `https://alexa.amazon.com/api/users/me` since as I've said before `.pl` variant doesn't exists. 

I'm wondering that maybe it's related to the fact that there is amazon officially in Poland but, I think, alexa isn't officially supported there.

Thanks, for your prompt reply and please let me know if I can help in any way.

maxmati:
I've contacted amazon support and they've confirmed that my settings are correct and expected in Poland (`countryOfResidence = PL` and  `marketPlaceDomainName = https://www.amazon.com/`). This because there is no digital content market nor official Alexa support in Poland, despite there is https://www.amazon.pl with physical market. 

I've also checked on my friends account and he have the same values.

src/aioamazondevices/const.py
```diff
@@ -71,6 +71,7 @@
         "domain": "com.au",
         "openid.assoc_handle": f"{DEFAULT_ASSOC_HANDLE}_au",
     },
+    "pl": AMAZON_US_OVERRIDE | {"market": "https://www.amazon.com"},
```
chemelli74, line 74:
```suggestion
    "pl": AMAZON_US_OVERRIDE,
```

You are specifying the default as exception; is not necessary

chemelli74 (changes requested):

src/aioamazondevices/const.py
maxmati: As I've mentioned in one of previous comments it doesn't work for me without overwriting market. There is an error that I get:
```
Selected country <PL> doesn't matches Amazon API reply:
{'input ': ['https://www.amazon.com']}
 vs 
{'amazon': 'https://www.amazon.com'}
```

chemelli74: This is another issue and has nothing to do with this PR.

chemelli74: I just tirggered a "update branch" so this issue is fixed.

Note: your commit message is not commitlint compliant, you need to fix it ;-)

chemelli74:
Thx for your contribution !

PR #467, opened 2025-09-01
fix: standardise locale

jamesonuk:
Current code results in `en-UK` (region is deprecated in `langcodes` and returns `territory` which is what is being passed in to start with).

UK is probably the only country where domain <> ISO country code so not sure it it is simpler to treat this as we do matching `com` to US ??


chemelli74:
Thx for spotting, comes from the change country->domain ;-)

Tested and works.

chemelli74 (approved):

PR #468, opened 2025-09-01
feat: alexa info skills

jamesonuk:
Add the inbuilt Alexa routine actions.

Not sure whether we want these in `AmazonSequenceType` enum for completeness or just in `ALEXA_INFO_SKILLS`

### DIFF
--- a/src/aioamazondevices/api.py
+++ b/src/aioamazondevices/api.py
@@ -112,23 +112,6 @@ class AmazonSequenceType(StrEnum):
     Music = "Alexa.Music.PlaySearchPhrase"
     TextCommand = "Alexa.TextCommand"
     LaunchSkill = "Alexa.Operation.SkillConnections.Launch"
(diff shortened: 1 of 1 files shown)


chemelli74:
I'm strongly against redundancy. 

I vote to create a AmazonInfoSkill enum and list them only there

jamesonuk:
> I vote to create a AmazonInfoSkill enum and list them only there

An enum makes `elif message_type in ALEXA_INFO_SKILLS:` a bit of a mess
A list is much simpler for this.   I have removed duplication from `AmazonSequenceType`
Is this better now?

What I wasn't sure of if the HA side.   I guess this would be a separate service with a drop down list of values (`ALEXA_INFO_SKILLS`) and how that would get translated.    Does this structure work for that?

chemelli74 (approved):
Nice one, thx !

PR #474, opened 2025-09-04
fix: consider errors from sensors

chemelli74:


jamesonuk (approved):
This has been running on my system for a while without issue.
Only real change is we no longer filter out only `RETRIEVABLE` properties and have added an `error` flag.
From a library point of view this is better.    The usefulness of this will be in integration changes

PR #487, opened 2025-09-13
fix: align sensor logic to new Amazon query output

jamesonuk:


src/aioamazondevices/const.py
```diff
@@ -77,7 +77,7 @@
         "scale": "scale",
     },
     "motionSensor": {
-        "name": "motion",
+        "name": "detectionState",
```
jamesonuk, line 80: just realised this is used as sensor name.   So changing it either needs change HA side or are we better factoring to use the key above as sensor name to pass to HA.    The new value of `detectionState` could be used be other sensors so isn't ideal

PR #495, opened 2025-09-17
feat: get supported music providers

jamesonuk:
## Summary by CodeRabbit

* **New Features**
  * Devices now discover and list available music services and indicate the default provider.
  * A new property exposes provider metadata (name, availability, default designation).

* **Refactor**
  * Music playback now selects services by provider ID (string) instead of fixed constants.
  * Playback validates provider availability and refreshes the provider list during device data updates to reduce invalid selections.


chemelli74 (dismissed):
A small change

src/aioamazondevices/api.py
```diff
@@ -204,11 +210,16 @@ async def call_alexa_music(
         self,
         device: AmazonDevice,
         search_phrase: str,
-        music_source: AmazonMusicSource,
+        provider_id: str,
     ) -> None:
         """Call Alexa.Music.PlaySearchPhrase to play music."""
+        if self._media_handler.music_providers is None:
+            raise ValueError("Music providers have not been initialised.")
```
chemelli74: This should not be needed anymore now that we update data on init, right ?

src/aioamazondevices/implementation/media.py
```diff
@@ -27,6 +33,12 @@ def __init__(
         """Initialize AmazonMediaHandler class."""
         self._session_state_data = session_state_data
         self._http_wrapper = http_wrapper
+        self._music_providers: dict[str, AmazonMusicProvider] | None = None
+
+    @property
+    def music_providers(self) -> dict[str, AmazonMusicProvider] | None:
+        """Return music providers."""
+        return self._music_providers
```
chemelli74: We may remove `None` as return type and just call the update if _self._music_provers_ is `None`

src/aioamazondevices/api.py
jamesonuk: It was defensive and I think to please pylance or mypy about something potentially not been initialised.

It should be safe to remove but may need something to check it as it is set to None on creation

src/aioamazondevices/implementation/media.py
chemelli74: Done, please check my latest commit

src/aioamazondevices/api.py
chemelli74: Remove `None` type, please check my latest commit.

chemelli74 (approved):
Good to go for me !

PR #501, opened 2025-09-21
refactor: cleanup and optimize

chemelli74:


src/aioamazondevices/api.py
```diff
@@ -1171,9 +1158,11 @@ async def _refresh_data(self, data_type: str) -> tuple[bool, dict]:
             "domain": f"www.amazon.{self._domain}",
         }
 
-        raw_resp = await self._session.post(
```
jamesonuk, line 1174: I can't remember the details but ISTR this was using `_session.post` and not `_session.request` because something was being overridden in or set in the request call that was causing this not to work.

If we can register a new device then it is probably fine but needs a bit of testing

chemelli74: Yeah agree, I did a single test from scratch but one more from your side would be better.

jamesonuk: Seems to work fine (tested with a couple of new registrations).   Only concern is whether that `x-amzn-identity-auth-domain` header is needed in any region but certainly here is working fine.

jamesonuk (approved):

PR #506, opened 2025-09-26
feat: comms and announcement enablement

### LINKED ISSUE #450: Announcements do not check whether device allows them
Within the Alexa app it is possible to disable announcements on individual devices.

We still create an active announcement entity and allow sending of announcements.   This all seems to register as worked however no announcement is made (the request is just ignored amazon side but returned as OK (200) respsone.

Need to look at how we identify this setting (do you need a switch in HA?) and take it into account when making announcements (just log, raise exception?)

jamesonuk:
Fixes #450

See initial commit as this requires a separate call per device but there is an endpoint
` f"https://{COMM_SITE}/homegroups/{self._home_group_id}/devices`
which returns all the data in one go however the `announcementAvailability` entity doesn't update in there

## Summary by CodeRabbit

* **New Features**
  * Device communication preferences are now accessible and displayed for each device.
  * Added controls to manage per-device communication status with enable/disable options.
  * Added drop-in status management with three configurable settings: All, Home, or Off.
  * Added ability to control announcement enablement on individual devices.
  * Communication settings are automatically fetched and integrated into device information during discovery.


src/aioamazondevices/api.py
```diff
@@ -1135,6 +1171,26 @@ async def set_do_not_disturb(self, device: AmazonDevice, state: bool) -> None:
...
+    async def set_announcements_enablement(
+        self, device: AmazonDevice, state: bool
+    ) -> None:
+        """Enable / disable announcements for device."""
+        payload = {"state": "ON" if state else "OFF"}
+        url = f"https://{COMM_SITE}/devicesTypes/{device.device_type}/deviceId/{device.serial_number}/preferences/announcements"
+        await self._session_request(
+            method=HTTPMethod.PATCH, url=url, input_data=payload, json_data=True
+        )
+
```
chemelli74, line 1193: Seems we can have a common pattern here

src/aioamazondevices/api.py
```diff
@@ -877,6 +911,8 @@ async def get_devices_data(
                 else None,
                 endpoint_id=device_endpoint["endpointId"] if device_endpoint else None,
                 sensors=sensors,
+                comms_enabled=device_comms_prefs.get("communications", "n/a"),
+                announcements_enabled=device_comms_prefs.get("announcements", "n/a"),
```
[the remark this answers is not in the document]
jamesonuk: drop-in isn't bool it has
```
      "allowedStates": [
        "All",
        "Home",
        "Off"
      ]
 ```
 
Also not sure we want calling and messaging.   I cannot see anywhere in the app where these can be set.   There is a global setting for calling but I can't seem to change anything and it isn't at device level.

I have redone this to just return the preference and state as `dict[str,str]`.   Library then just returns what API says and we can apply the logic to use which we want on HA side

File: src/aioamazondevices/implementation/communication.py [hunk 3]
```diff
@@ -0,0 +1,74 @@
...
+        await self._set_communications_state(
+            "announcements", device, "ON" if state else "OFF"
+        )
+
+    async def set_dropin_enablement(self, device: AmazonDevice, state: str) -> None:
+        """Set allowed dropin state for device.
+
+        State values are All, Home and Off
+        """
+        await self._set_communications_state("announcements", device, state)
```
chemelli74: Not sure **drop-in** needs the preference `"announcements"`

chemelli74 (commented):
A couple open comments and then we should be good to go with this one as well ;-)

src/aioamazondevices/implementation/communication.py
```diff
@@ -0,0 +1,74 @@
...
+    ) -> None:
+        """Enable / disable announcements for device."""
+        await self._set_communications_state(
+            "announcements", device, "ON" if state else "OFF"
+        )
+
+    async def set_dropin_enablement(self, device: AmazonDevice, state: str) -> None:
+        """Set allowed dropin state for device.
+
+        State values are All, Home and Off
```
chemelli74: We should valide them or use a StrEnum like we did for all scenarios

src/aioamazondevices/implementation/communication.py
```diff
@@ -0,0 +1,74 @@
...
+        """
+        await self._set_communications_state("dropin", device, state)
+
+    async def get_communication_preferences(
+        self, devices: list[AmazonDevice]
+    ) -> dict[str, dict[str, str]]:
+        """Get communication preferences for a device."""
+        communication_preferences = {}
+        for device in devices:
+            url = f"https://{COMM_SITE}/devicesTypes/{device.device_type}/deviceId/{device.serial_number}/preferences?devicePreferences=communications&devicePreferences=calling&devicePreferences=messaging&devicePreferences=dropin&devicePreferences=announcements"
```
chemelli74: Let's switch to URL() from yarl and use the query syntax to build it.
Much more clear and error prone.

Note: Something to do on all urls in the library.

src/aioamazondevices/implementation/communication.py
```diff
@@ -0,0 +1,74 @@
...
+            _, resp = await self._http_wrapper.session_request(
+                method=HTTPMethod.GET, url=url
+            )
+            resp_json = await self._http_wrapper.response_to_json(resp)
+
+            comms_preferences = {}
+            for device_prefs in resp_json.get("devicePermissionsPreferences", {}):
+                comms_preferences[device_prefs.get("devicePreference")] = (
+                    device_prefs.get("state", "n/a")
```
chemelli74: Can we go with `None` instead of `"n/a"` ?

chemelli74 (changes requested):

src/aioamazondevices/const/http.py
```diff
@@ -57,3 +58,5 @@
 HTTP2_SITE = "https://bob-dispatch-prod-{region}.amazon.com"
 HTTP2_DIRECTIVES_VERSION = "20160207"
 HTTP2_RECONNECT_DELAY = 5  # seconds
+
+COMM_SITE = "alexa-comms-mobile-service.amazon.com"
```
chemelli74:
```suggestion
COMM_SITE = "https://alexa-comms-mobile-service.amazon.com"
```

src/aioamazondevices/const/http.py
```diff
@@ -37,6 +37,7 @@
 URI_CAPABILITIES = "v1/devices/@self/capabilities"
 URI_BEHAVIORS_PREVIEW = "api/behaviors/preview"
 URI_BEHAVIORS_AUTOMATIONS = "api/behaviors/v2/automations"
+URI_COMM_PREFERENCES = "devicesTypes/{device_type}/deviceId/{serial_number}/preferences"
```
chemelli74, line 40: Note: we should use the same approach with format() all over the library (es. todo.py)

src/aioamazondevices/api.py
```diff
@@ -472,6 +485,24 @@ async def set_do_not_disturb(self, device: AmazonDevice, enable: bool) -> None:
         """Set Do Not Disturb status for a device."""
         await self._dnd_handler.set_do_not_disturb(device, enable)
 
+    async def set_communication_enabled(
```
```suggestion
    async def set_communication(
```

Like we did here: <https://github.com/chemelli74/aioamazondevices/blob/6a6364f8d71775b3291cc1e315d79fb203b326ff/src/aioamazondevices/api.py#L477>

jamesonuk: My issue is you can set do not disturb and it makes sense.
You can't really set communications and know what that is going to do.
It is more announcements where `set_announcement` sounds more like setting the contents of an announcement rather than enabling / disabling.

I think not having something to indicate that it is enabling might be confusing.

That said I am not precious.   If you really want
* `set_communication`
* `set_announcement`
* `set_drop_in`

I can make that so

chemelli74: You are correct. Probably best will be:

- set_communications_state
- set_announcements_state
- set_dropin_state

Let me know if those work (from a native English speaking 😉

chemelli74 (approved):
Nice, thx !

PR #508, opened 2025-09-28
chore: remove sporadic log warning

chemelli74:


jamesonuk (approved):
Tested and removes the logging when response is empty.

PR #518, opened 2025-10-05
fix: improve owner ID retrieval logic

chemelli74:
## Summary by CodeRabbit

- Refactor
  - Switched device processing to use a local devices map with lazy owner-id lookup to reduce shared mutable state.

- Bug Fixes
  - Improved owner-id retrieval logic with clearer error handling when the account owner cannot be determined.

- Chores
  - Bumped release to v6.2.9-rc.1 and added a pre-release changelog entry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## v6.2.9-rc.1 (2025-10-05)
+
+### Bug fixes
+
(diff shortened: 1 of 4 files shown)


jamesonuk (approved):
OK so with main announcement / speak fails with bad request (customer ID is null in payload).

In branch in works fine and it debug logs account ID
`2025-10-05 14:07:33.278 DEBUG (MainThread) [aioamazondevices] Setting account owner`

PR #552, opened 2025-10-21
fix: add filters by capabilities

chemelli74:
## Summary by CodeRabbit

* **Bug Fixes**
  * Missing sensors are no longer created when reported as not found, preventing phantom sensor entries.
  * Do Not Disturb and other sensors are skipped for multi-speaker/group devices to avoid incorrect sensor attachments.
  * Devices without microphone/voice capability are excluded from voice device lists so non-voice devices no longer appear where voice devices are expected.

* **Refactor**
  * Device setup logic clarified by centralizing account and capability handling for more reliable initialization.

### DIFF
--- a/src/aioamazondevices/api.py
+++ b/src/aioamazondevices/api.py
@@ -40,6 +40,7 @@ from .const import (
     AMAZON_CLIENT_OS,
     AMAZON_DEVICE_SOFTWARE_VERSION,
     AMAZON_DEVICE_TYPE,
+    ARRAY_WRAPPER,
     BIN_EXTENSION,
     COUNTRY_GROUPS,
     CSRF_COOKIE,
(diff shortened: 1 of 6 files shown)


src/aioamazondevices/api.py
```diff
@@ -1153,7 +1154,9 @@ async def _get_sensor_data(self) -> None:
             else:
                 for device_sensor in device.sensors.values():
                     device_sensor.error = True
-            if device_dnd := dnd_sensors.get(device.serial_number):
+            if (
+                device_dnd := dnd_sensors.get(device.serial_number)
+            ) and device.device_family != SPEAKER_GROUP_FAMILY:
```
jamesonuk: I was going to say that if they don't have a DND they shouldn't be in the DND response but after testing I see that speaker groups do return a DND status but it doesn't appear to reflect the children (ie. it isn't on if any/all of the speakers that make up that group have DND set).   In the app I can see no way to get or set this value so 

The DND endpoint is also returning a value for an old Sonos device I have but again I cannot see anyway to set DND for this.   Just looking to see if there is any other way to filter out devices that don't actually have a DND (there is no capability)

francescolf (approved):
Working as expected with 3 Echo devices and a TV

francescolf (approved):

PR #558, opened 2025-10-23
refactor: remove save_to_file() method

chemelli74:
Small refactor to make #554 smaller

## Summary by CodeRabbit

* **Removed Features**
  * The `--save_raw_data` CLI option has been removed.
  * Legacy internal raw-saving toggle has been removed.

* **Documentation**
  * Configuration examples updated to remove the `save_raw_data` entry.

* **New Features**
  * Adds a ".raw" file extension for raw output files.
  * Introduces a customizable asynchronous save handler for flexible, extensible persistence of output data.

### DIFF
--- a/src/aioamazondevices/api.py
+++ b/src/aioamazondevices/api.py
@@ -404,14 +404,16 @@ class AmazonEchoApi:
                     f"Request failed: {await self._http_phrase_error(resp.status)}"
                 )
 
+        raw_content = await resp.read()
+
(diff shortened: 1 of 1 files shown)


chemelli74:
> Seems to work fine. Just wondering whether save to file should be part of library or not. Could this be done by API accepting a coroutine for saving output and if present it gets run where we save to file. So save to file definition would only be in library test script?

Why not, let's draft this for now.

Issue #537, opened 2025-10-12
Amazon Air Quality Monitor


cristianomeda:
Would it be possible to add the support for the Amazon Air Quality Monitor? Thank you

PR #573, opened 2025-10-27
feat: add support for aqm devices

jamesonuk:
addresses https://github.com/chemelli74/aioamazondevices/issues/537
would need logic embedding on HA side to convert `rangeValue-4` into `humidity`
`rangeValue-6` into `PM25` etc. and add UoM etc.

There are six additional sensors and some of these have ranges
eg. for VOC 0-25 = good, 26-65 = moderate and 66-100 = poor
not sure if we want to replicate those textual descriptions as well as the raw values

## Summary by CodeRabbit

* **New Features**
  * Air Quality Monitor (AQM) devices are now discovered and listed alongside other devices.
  * New range-based sensors (humidity, PM2.5 and related metrics) are shown with proper names and units so readings appear correctly in device details and sensor lists.

* **Tests**
  * Test utility updated to print AQM device details and their sensors for verification.

### DIFF
--- a/library_test.py
+++ b/library_test.py
@@ -16,7 +16,7 @@ from aiohttp import ClientSession
 from colorlog import ColoredFormatter
 
 from aioamazondevices.api import AmazonEchoApi
-from aioamazondevices.const.metadata import AQM_DEVICE_TYPE
+from aioamazondevices.const.devices import AQM_DEVICE_TYPE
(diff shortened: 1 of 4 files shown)


chemelli74 (approved):
Nice one, thx !

Note: cannot test due to lasck of device

PR #617, opened 2025-11-20
fix: handle rate limit errors on notification endpoint

jamesonuk:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling when retrieving notifications so failures no longer interrupt operation; missing notification data is now handled gracefully.
  * When notification data cannot be obtained, the system skips notification updates and continues normal processing without raising errors.


chemelli74 (approved):
LGTM now, thx !

PR #625, opened 2025-11-25
feat: store account customer id

jamesonuk:
https://github.com/chemelli74/aioamazondevices/pull/503 switched from using the device owner ID (from restful devices endpoint) to use the account customer ID so that speech and announcements would work with household devices.
This account customer ID was looked up by matching the serial number of registered aioamazondevices device to use that in all calls.    This had several subsequent PRs as it took some time for the device to appear in the API responses and we have ended up retrying up to 30 times with a two second delay and this is set to happen at every start up.

This ID is however available on the `api/users/me` endpoint without any delay.

This PR will do the lookup once and then add it to the stored login data (for new HA installs that will persist in HA config) for existing configs it will still do the lookup once each start up but a HA PR will address that and store it HA side,

## Release Notes

* **Improved Login**
  * Streamlined account identification during sign-in.
  * Login now avoids repeating account lookups when account information is already available.
  * Account details are retrieved directly from the current user service for a simpler, more efficient login flow.

* **Refactor**
  * Consolidated account information handling to improve consistency across session data.


chemelli74:
@coderabbitai, full review please

chemelli74 (approved):
Thx for the review, please add some more details to this PR description and then we can merge.

PR #626, opened 2025-11-25
refactor: make info skills a dict to allow for translation

jamesonuk:
This is needed as translation keys in HA need to be lower case with no dots

## Summary by CodeRabbit

* **Refactor**
  * Standardized internal Alexa skill references to a stable mapping, improving consistency and message resolution for Alexa-related features with no change to user-visible behavior.

* **Bug Fix**
  * Corrected the Alexa info skill identifier used for single-device requests to ensure the proper skill is invoked.


src/aioamazondevices/api.py
```diff
@@ -487,7 +487,8 @@ async def call_alexa_info_skill(
         info_skill_name: str,
```
```suggestion
        info_skill_name: list[ALEXA_INFO_SKILLS],
```

or something like that so you don't have to check if the parameter is correct

PR #643, opened 2025-12-22
fix: user agent

chemelli74:
## Summary by CodeRabbit

* **Improvements**
  * Enhanced HTTP request handling by implementing configurable User-Agent headers across API communications.
  * Added debug logging for extended headers to improve troubleshooting and diagnostics capabilities.

### DIFF
--- a/src/aioamazondevices/http_wrapper.py
+++ b/src/aioamazondevices/http_wrapper.py
@@ -289,11 +289,13 @@ class AmazonHttpWrapper:
         )
 
         headers = DEFAULT_HEADERS.copy()
-        headers.update({"User-Agent": REQUEST_AGENT["Amazon"]})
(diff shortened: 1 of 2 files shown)


src/aioamazondevices/http_wrapper.py
```diff
@@ -289,11 +289,13 @@ async def session_request(
         )
 
         headers = DEFAULT_HEADERS.copy()
-        headers.update({"User-Agent": REQUEST_AGENT["Amazon"]})
+        headers.update({"User-Agent": REQUEST_AGENT["Browser"]})
```
jamesonuk: I am still not sure why we are defaulting to a browser user-agent.   All these calls reflect the app and that does not use a browser user-agent.    I think this works for notifications only because the user-agent is different to what is there now not because it should be this.    I am fairly confident that this will just end up being rate limited as the current user-agent is now.

PR #648, opened 2025-12-28
fix: add connectivity to graphql sensors

### LINKED ISSUE #647: Connectivity Status is only updated once per day
Currently we set the `online` attribute for a device based on the `URI_DEVICES` call that takes place at start up and then once every 24 hours.

We could possibly add the graphQL entity back now that it seems stable (`Reachability` is already in the query but there needs to be a little checking on status, eg. I know a fire TV stick I have that is not connected is returning an error of `NOT_FOUND` which should really be treated as offline rather than unavailable I think but there was also a status for online.

I have `reachabilityStatusValue` of `OK`, `UNREACHABLE` (on a fire tablet where battery has ran out recently) and null with the `NOT_FOUND` error on devices that are either logical (Alexa Simulator) or have been offline so long there is no cached history

jamesonuk:
Fixes #647

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved device online-status detection using connectivity sensor data.
  * Corrected online-status calculation for speaker groups (group marked online only when all members are online).

* **New Features**
  * Added a display listing all devices with account names and their online status.

### DIFF
--- a/src/aioamazondevices/api.py
+++ b/src/aioamazondevices/api.py
@@ -16,6 +16,7 @@ from .const.devices import (
 from .const.http import (
     ARRAY_WRAPPER,
     DEFAULT_SITE,
+    REQUEST_AGENT,
     URI_DEVICES,
     URI_NEXUS_GRAPHQL,
 )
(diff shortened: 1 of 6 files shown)


src/aioamazondevices/api.py
```diff
@@ -216,7 +216,10 @@ def _get_device_sensor_state(
                         "error in sensor %s - %s - %s", name, error_type, error_msg
                     )
 
-                if error_type != "NOT_FOUND":
+                if (
+                    error_type != "NOT_FOUND"
+                    or feature_property.get("name") == "reachability"
```
chemelli74: why do we need this check ?

jamesonuk: I think this was a hangover from a previous iteration before I set the else to `online = False`
(without out there is no reachability sensor created)
Will double check but I think it can likely be removed

PR #749, opened 2026-03-30
refactor: internal method to clarify refresh job

chemelli74:
## Summary by CodeRabbit

* **New Features**
  * Device status, Do Not Disturb, notifications and sensor data are now refreshed automatically before updates are fetched.

* **Improvements**
  * Daily refresh now also updates the base device list and routines for more up-to-date behavior.
  * Refresh throttling refined to reduce redundant calls while preserving endpoint-specific throttling.


jamesonuk (approved):
This seems better now to avoid breaking external API

PR #783, opened 2026-04-11
refactor: use devices endpoint for volume not sequence

jamesonuk:
This might be needed to avoid rate limiting issues using sequence/routines to set volume.
Need to just check how this works with speaker groups (and stereo pairs)


chemelli74:
Should we get volumes only once at startup and then refresh only via HTTP2 push ?

jamesonuk:
> Should we get volumes only once at startup and then refresh only via HTTP2 push ?

This is setting volume not getting.

src/aioamazondevices/implementation/media.py
```diff
@@ -35,17 +36,40 @@ async def get_device_volumes(self) -> dict[str, AmazonVolumeState]:
...
+            self._device_volumes[device_volume_data["dsn"]] = AmazonVolumeState(
                 device_volume_data["speakerVolume"], device_volume_data["speakerMuted"]
             )
 
-        return _volumes
+        return self._device_volumes
+
+    async def set_device_volume(self, device: AmazonDevice, volume: int) -> None:
+        """Set device volume."""
+        if not self._device_volumes:
```
chemelli74, line 65: IF we get at init, this could not happen right ?

Same logic as per music_providers ;-)

jamesonuk: subtle check.
In init it is `self._device_volumes: dict[str, AmazonVolumeState] = {}` so empty dict.
so `if not self._device_volumes` will be true and this will cause the dict to be populated.

So essentially it is possible to call `set_device_volume` before `get_device_volumes`.

This method needs to know the current volume as it sends a delta value rather than a literal and this check ensures if `_device_volumes` is an empty array we try and load it.

src/aioamazondevices/implementation/media.py
```diff
@@ -49,17 +50,40 @@ async def get_device_volumes(self) -> dict[str, AmazonVolumeState]:
...
+            "dsn": device.serial_number,
+            "deviceType": device.device_type,
+            "amount": diff,
+            "volume": volume,
+            "muted": True,
+            "synchronous": True,
+        }
+        await self._http_wrapper.session_request(
+            method=HTTPMethod.PUT,
+            url=f"https://alexa.amazon.{self._session_state_data.domain}/api/devices/{device.device_type}/{device.serial_number}/audio/v2/speakerVolume",
```
chemelli74, line 83: Let's use yarl URL() here as well.

PR #784, opened 2026-04-11
refactor: parse using mime boundaries

jamesonuk:
for review about how http2 stream is being parsed.

There are a number of issues in the current logic.
* we expect one chunk to equal one message
* we are processing text not bytes
* we are manually parsing messages to establish boundaries and extract headers

This uses `BytesParser` from `email.parser` which despite the package name is purely a MIME parser specifically for multipart messages (which are really only widely seen in emails).

Not ready for merging but for review.   The SSL changes are probably moot as whilst there is a theoretical overhead to recreate this there shouldn't be any reconnections.   Similarly the one chunk = one message is holding true in the data I have seen but there is a good change that on a slower or problematic connection that this might not be the case.

This still also only process the first `renderingUpdates`.   We are only seeing one per message but given it is an array there is little issue with looping over all values rather than only picking the first (this is still an issue)

## Summary by CodeRabbit

* **Refactor**
  * Renamed `start_http2_thread()` to `start_http2_processing()` and `stop_http2_thread()` to `stop_http2_processing()`.
  * Updated `start_http2_processing()` to accept an HTTP client parameter for external management.
  * Improved internal directive stream handling and connection management.
  * Removed `is_connected()` method.


src/aioamazondevices/implementation/http2.py
```diff
@@ -26,6 +29,15 @@
 from aioamazondevices.structures import AmazonPushMessage
 from aioamazondevices.utils import _LOGGER
 
+_BOUNDARY_RE = re.compile(r'boundary="?([^";,]+)"?', re.IGNORECASE)
+
+
+def _parse_boundary(content_type: str) -> bytes | None:
+    match = _BOUNDARY_RE.search(content_type)
+    if not match:
```
chemelli74:
```suggestion
    if not (match := BOUNDARY_RE.search(content_type)):
```

src/aioamazondevices/implementation/http2.py
```diff
@@ -84,117 +84,155 @@ async def _register_device_capabilities(self) -> None:
...
+            await self._register_device_capabilities()
+
+        await self._http2_init_client()
+        return True
+
+    async def _stream_and_process(self) -> None:
+        """Open stream and process incoming directives."""
+        _LOGGER.debug("Starting AVS Directives stream")
+
+        if not self._http2_client:
```
chemelli74: This method is only internal, so we know it's called after http2_client init right ?

src/aioamazondevices/implementation/http2.py
```diff
@@ -84,117 +84,155 @@ async def _register_device_capabilities(self) -> None:
...
-                _LOGGER.debug("Device capabilities not set, registering now.")
-                await self._register_device_capabilities()
+            if response.status_code in (
+                HTTPStatus.UNAUTHORIZED,
+                HTTPStatus.FORBIDDEN,
+            ):
+                raise CannotAuthenticate
 
-            await self._http2_init_client()
+            boundary = _parse_boundary(response.headers.get("content-type", ""))
```
chemelli74: why not move the method in the class and make it self-consistent ?

src/aioamazondevices/implementation/http2.py
```diff
@@ -84,117 +84,155 @@ async def _register_device_capabilities(self) -> None:
...
-            if not self._stop_event.is_set():
-                _LOGGER.debug(
-                    "Reconnecting to HTTP2 endpoint in %s seconds", self.reconnect_delay
+            async for chunk in response.aiter_bytes():
+                if self._stop_event.is_set():
+                    break
+
+                buffer.extend(chunk)
+
+                if len(buffer) > 512 * 1024:  # 512KB limit to prevent memory issues
```
chemelli74: Please make it a const instead of using "magic numbers" in the code

src/aioamazondevices/implementation/http2.py
```diff
@@ -84,117 +84,155 @@ async def _register_device_capabilities(self) -> None:
...
+                    _LOGGER.error("Buffer exceeded maximum size, forcing reconnect")
```
chemelli74:
```suggestion
                    _LOGGER.error("Buffer of %s exceeded maximum size, forcing reconnect", AVS_STREAM_BUFFER_SIZE)
```

src/aioamazondevices/implementation/http2.py
```diff
@@ -84,117 +84,155 @@ async def _register_device_capabilities(self) -> None:
...
+                    del buffer[: idx + len(boundary)]
+
+                    await self._handle_part(part.strip())
+
+        _LOGGER.debug("AVS Directives stream closed")
+
+    async def _handle_part(self, part: bytes) -> None:
+        """Process a single multipart section."""
+        if not part:
+            await self._ping()
```
chemelli74: PRobably better to move to a parallel task; was desing like this for a PoC not for prod

src/aioamazondevices/implementation/http2.py
```diff
@@ -84,117 +84,155 @@ async def _register_device_capabilities(self) -> None:
...
+            return
+
+        chunk_json = self._extract_json_from_part(part)
+
+        try:
+            updates_nodes = chunk_json["directive"]["payload"]["renderingUpdates"]
+        except (KeyError, IndexError):
+            _LOGGER.warning("Malformed directive payload")
```
chemelli74:
```suggestion
            _LOGGER.warning("Malformed directive payload: %s", chunk_json)
```

src/aioamazondevices/implementation/http2.py
```diff
@@ -84,117 +84,155 @@ async def _register_device_capabilities(self) -> None:
...
+            return
+
+        # in practce all messages seem to contain only a single update
+        # but this is an array so loop over results
```
chemelli74:
```suggestion
        # All observed messages appear to contain only one update
        # but this is treated as an array to ensure we iterate over all results
```

src/aioamazondevices/implementation/http2.py
```diff
@@ -84,117 +84,155 @@ async def _register_device_capabilities(self) -> None:
...
+
+            # Skip duplicate NotificationChange pushes
+            if (
+                push_event_type == AmazonPushMessage.NotificationChange.value
+                and payload.get("notificationVersion", 2) % 2 == 0
+            ):
+                return
+
+            _LOGGER.debug(
+                "Detected push type <%s> on device <%s>",
```
chemelli74:
```suggestion
                "Detected push event type <%s> on device <%s>",
```

src/aioamazondevices/implementation/http2.py
```diff
@@ -207,60 +245,56 @@ def _string_recursive_parse(
...
-        # remove potential boundary terminator
-        if "\r\n--------" in body:
-            body = body.split("\r\n--------", 1)[0]
+        if msg.get_content_type() != "application/json":
+            raise ValueError("Unexpected content-type")
 
-        # parse top-level JSON
-        top_level_json = orjson.loads(body)
+        body = msg.get_payload(decode=True)
+        if body is None:
```
chemelli74:
```suggestion
        if (body := msg.get_payload(decode=True)) is None:
```

src/aioamazondevices/implementation/http2.py
```diff
@@ -207,60 +245,56 @@ def _string_recursive_parse(
...
 
     async def _http2_init_client(self) -> None:
         """Create HTTP2 client session."""
-        if hasattr(self, "_http2_client"):
+        if self._http2_client and not self._http2_client.is_closed:
             await self._http2_client.aclose()
 
-        ssl_context = await self._build_ssl_context_async()
+        if self._ssl_context is None:
+            self._ssl_context = await asyncio.to_thread(create_default_context)
```
chemelli74: why not assign in __init__ once for ever ?

src/aioamazondevices/implementation/http2.py
```diff
@@ -207,60 +245,56 @@ def _string_recursive_parse(
...
+        region = self._session_state_data.login_stored_data["customer_info"][
+            "home_region"
+        ]
+        return HTTP2_SITE.format(region=region)
 
     async def _ping(self) -> None:
         """Ping."""
-        if not self._login_stored_data:
-            _LOGGER.warning("No login data available, cannot get directives")
+        if not self._http2_client:
```
chemelli74: SAme as above, internal method s owe know it's called after http2_client init

src/aioamazondevices/implementation/http2.py
```diff
@@ -277,5 +311,11 @@ async def _ping(self) -> None:
                 "Detected ping 403, please check your credentials and region"
             )
 
-    async def _build_ssl_context_async(self) -> SSLContext:
-        return await asyncio.to_thread(create_default_context)
+    def _get_bearer_token(self) -> str:
+        """Get current bearer token."""
+        if not (
+            token := self._session_state_data.login_stored_data.get("access_token")
```
chemelli74: I think we used const in other places, so better to make it consistent

chemelli74 (changes requested):
A few minor things ;-)

src/aioamazondevices/implementation/http2.py
jamesonuk: done

src/aioamazondevices/implementation/http2.py
jamesonuk: Is there really any value in knowing the maximum size?   This is of now use for users and of little use for debugging as it is was it is.   This is purely to catch a scenario where somehow we keep getting the http/2 stream with no boundaries so the buffer just grows forever and will eventually kill the system with memory issues.   Very unlikely but I am not sure there is much value in knowing the buffer size?

src/aioamazondevices/implementation/http2.py
jamesonuk: Yes indeed, that is the one last thing I want to resolve

src/aioamazondevices/implementation/http2.py
jamesonuk: done

src/aioamazondevices/implementation/http2.py
jamesonuk: done

src/aioamazondevices/implementation/http2.py
jamesonuk: done

src/aioamazondevices/implementation/http2.py
jamesonuk: token isn't a constant though ?

src/aioamazondevices/implementation/http2.py
jamesonuk: done

src/aioamazondevices/implementation/http2.py
jamesonuk: This was to please mypy

src/aioamazondevices/implementation/http2.py
jamesonuk: done, but is only used in one place

src/aioamazondevices/implementation/http2.py
```diff
@@ -84,117 +89,180 @@ async def _register_device_capabilities(self) -> None:
...
 
-            if not self._login_stored_data.get(DEVICE_CAPABILITIES_REGISTERED, False):
-                _LOGGER.debug("Device capabilities not set, registering now.")
-                await self._register_device_capabilities()
+            try:
+                await self._stream_and_process()
+            except httpx.RemoteProtocolError as exc:
+                _LOGGER.debug("HTTP2 disconnection detected: %s", exc)
+            except httpx.HTTPError as exc:
+                _LOGGER.warning("HTTP2 error detected: %s", exc)
```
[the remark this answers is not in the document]
jamesonuk: @coderabbitai with these exceptions retrying will not resolve them.   There is no point keeping loop alive as every loop will fail and external changes need to be made that will involve restarting the loop

src/aioamazondevices/implementation/http2.py
chemelli74: "access_token"

jamesonuk: In two minds here, this appears to be the value we are using a constant for when parsing JSON.

We do have https://github.com/chemelli74/aioamazondevices/blob/57eead09faabc16819168ec7bbfac67d5a5f3bc6/src/aioamazondevices/const/http.py#L33
which is used
https://github.com/chemelli74/aioamazondevices/blob/1be7b3befc85f30cacf77196f90d053169dd8f67/src/aioamazondevices/http_wrapper.py#L266
but we also have 
https://github.com/chemelli74/aioamazondevices/blob/1be7b3befc85f30cacf77196f90d053169dd8f67/src/aioamazondevices/login.py#L224
https://github.com/chemelli74/aioamazondevices/blob/1be7b3befc85f30cacf77196f90d053169dd8f67/src/aioamazondevices/login.py#L240
where we aren't using the constant (and would possibly be more confusing to do so).
but we are using this elsewhere so will apply it.

PR #800, opened 2026-04-21
chore: improve error handling after redirect to signin page

jamesonuk:
## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced error messaging for authentication failures — the app now displays a clearer "Authentication failed, check credentials" message when sign-in attempts fail, providing better guidance for resolving authentication issues.

### DIFF
--- a/src/aioamazondevices/http_wrapper.py
+++ b/src/aioamazondevices/http_wrapper.py
@@ -410,7 +410,7 @@ class AmazonHttpWrapper:
                 _LOGGER.debug("JSON '%s' data: %s", description, scrub_fields(data))
             return cast("dict[str, Any]", data)
         except ContentTypeError as exc:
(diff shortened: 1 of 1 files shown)


src/aioamazondevices/http_wrapper.py
```diff
@@ -410,6 +410,12 @@ async def response_to_json(
                 _LOGGER.debug("JSON '%s' data: %s", description, scrub_fields(data))
             return cast("dict[str, Any]", data)
         except ContentTypeError as exc:
+            if raw_resp.url.path == "/ap/signin":
```
chemelli74: This should be taken from a var in cost/http.py instead of writing here the path hardcoded

jamesonuk: done

chemelli74 (approved):

PR #820, opened 2026-05-08
fix: enhance logging

chemelli74:
## Summary by CodeRabbit

* **Chores**
  * Enhanced logging around media and volume state updates to improve observability when state signals are frozen.
  * Improved HTTP/2 error reporting to include detailed exception information for easier diagnosis of streaming issues.
  * Added debug logs for duplicate notification checks and for cases where a rendering update could not be processed, aiding troubleshooting.

[Review Change Stack](https://app.coderabbit.ai/change-stack/chemelli74/aioamazondevices/pull/820)


src/aioamazondevices/implementation/http2.py
```diff
@@ -115,6 +115,12 @@ def _process_rendering_update(  # noqa: PLR0911
         return None
 
     if _is_duplicate_notification(push_event_type, payload):
+        _LOGGER.debug(
+            "Ignoring duplicate push '%s' for device %s: %s",
+            push_event_type,
+            device_serial,
+            payload,
+        )
```
jamesonuk: This is the only one I would question.
It can only be a duplicate if it is a `NotificationChange` with an even `notificationVersion`
This is obviously future proof but not sure how much value there is in logging this (particularly the type /payload).

jamesonuk (approved):
Only one comment but not fussed either way on that one.   Just think that log would slightly bloat logs unnecessarily.    Just to consider, happy if you do want it in

src/aioamazondevices/implementation/http2.py
chemelli74: You are right.
Changed approach on my latest commit.

src/aioamazondevices/implementation/http2.py
```diff
@@ -61,6 +61,7 @@ def _is_duplicate_notification(push_event_type: str, payload: dict[str, Any]) ->
     Observed behavior: even notificationVersion values are duplicates.
     """
     if push_event_type != AmazonPushMessage.NotificationChange.value:
+        _LOGGER.debug("Not checking for duplicates on push type <%s>", push_event_type)
```
jamesonuk: getting to the point of rolling in glitter but if notifications are only 1 out of 100 or 1,000 events then there is a lot of logging telling us we aren't doing anything.

As this function is already called `_is_duplicate_notification` why not change the caller from
```python
if _is_duplicate_notification(push_event_type, payload):
```
to
```python
if push_event_type == AmazonPushMessage.NotificationChange.value and _is_duplicate_notification(payload):
```
and cut this leg out completely or if we want the defensive check raise an error here rather than logging?

This is fine as is but if we only call this for the events we actually want to check it would reduce logging

jamesonuk (approved):
As before, what is there is perfectly fine but suggestion might be to tighten the caller so we are aren't even checking for duplicates on other event types so no need to log.

src/aioamazondevices/implementation/http2.py
chemelli74: Done, much more clean !

jamesonuk (approved):

PR #839, opened 2026-05-18
feat: implement ToDo lists

lonlazer:
This PR adds ToDo list capabilities to aioamazondevices.

<details>
<summary>Example usage script</summary>

```python
import argparse
import asyncio
import json
import logging
import sys
from typing import Any, cast

from aiohttp import ClientSession
from anyio import Path

from aioamazondevices.structures import ListItem
from src.aioamazondevices import AmazonEchoApi

logging.basicConfig(level=logging.INFO, format="%(message)s")
logger = logging.getLogger(__name__)


async def read_from_file(data_file: str) -> dict[str, Any]:
    """Load stored login data from file."""
    if not data_file:
        return {}

    file_path = Path(data_file)
    if not await file_path.exists():
        logger.warning("Cannot find previous login data file: %s", data_file)
        return {}

    async with await file_path.open("rb") as f:
        content = await f.read()
        return cast("dict[str, Any]", json.loads(content))


def print_todo_list(list_items: list[ListItem]) -> None:
    """Print the todo list."""
    logger.info("Current To-Do List:")
    for list_item in list_items:
        logger.info("  - %s (%s)", list_item.name, list_item.status)
    logger.info("")


def get_item_by_name(list_items: list[ListItem], name: str) -> ListItem:
    """Find a list item by its original name."""
    try:
        return next(item for item in list_items if item.original_name == name)
    except StopIteration:
        logger.error("Could not find item with name: '%s'", name)
        sys.exit(1)


async def refresh_and_print(api: AmazonEchoApi, list_id: str) -> list[ListItem]:
    """Helper to wait, refresh the list, and print it."""
    await asyncio.sleep(2)
    await api.update_todo_list_items(list_id=list_id)
    items = api.todo_list_items[list_id]
    print_todo_list(items)
    return items


async def main() -> None:
    parser = argparse.ArgumentParser(description="Test Amazon ToDo API functionality.")
    parser.add_argument("email", help="Amazon account email")
    parser.add_argument("password", help="Amazon account password")
    parser.add_argument(
        "--login-data",
        help="Path to stored login data JSON file (default: %(default)s)",
    )
    args = parser.parse_args()

    login_data_stored = await read_from_file(args.login_data)

    async with ClientSession() as client_session:
        api = AmazonEchoApi(
            client_session=client_session,
            login_email=args.email,
            login_password=args.password,
            login_data=login_data_stored,
        )

        logger.info("Logging in...")
        await api.login.login_mode_stored_data()

        logger.info("Updating lists...")
        await api._todo_handler.update_lists()
        await api.update_todo_list_items()

        if not api.todo_lists:
            logger.error("No to-do lists found on the account.")
            return

        first_list_id = api.todo_lists[0].id

        # Initial List Display
        list_items = api.todo_list_items[first_list_id]
        print_todo_list(list_items)
        await asyncio.sleep(2)

        # 1. Add Item
        logger.info(">>> Action: Adding Item 'Cheese'")
        await api.add_todo_list_item(first_list_id, "Cheese")
        list_items = await refresh_and_print(api, first_list_id)
        cheese_item = get_item_by_name(list_items, "Cheese")

        # 2. Rename Item
        logger.info(">>> Action: Renaming Item 'Cheese' to 'Cheddar'")
        await api.rename_todo_list_item(
            first_list_id, cheese_item.id, "Cheddar", cheese_item.version
        )
        list_items = await refresh_and_print(api, first_list_id)
        cheddar_item = get_item_by_name(list_items, "Cheddar")

        # 3. Mark as Completed
        logger.info(">>> Action: Setting Item 'Cheddar' to completed")
        await api.set_todo_list_item_checked_status(
            first_list_id, cheddar_item.id, True, cheddar_item.version
        )
        list_items = await refresh_and_print(api, first_list_id)
        cheddar_item = get_item_by_name(list_items, "Cheddar")

        # 4. Mark as Incomplete
        logger.info(">>> Action: Setting Item 'Cheddar' to incomplete again")
        await api.set_todo_list_item_checked_status(
            first_list_id, cheddar_item.id, False, cheddar_item.version
        )
        list_items = await refresh_and_print(api, first_list_id)
        cheddar_item = get_item_by_name(list_items, "Cheddar")

        # 5. Delete Item
        logger.info(">>> Action: Deleting Item 'Cheddar'")
        await api.delete_todo_list_item(
            first_list_id, cheddar_item.id, cheddar_item.version
        )
        await refresh_and_print(api, first_list_id)

        logger.info("Test completed successfully.")


if __name__ == "__main__":
    asyncio.run(main())

```
</details>

## Summary by CodeRabbit

* **New Features**
  * Added comprehensive ToDo list management to the API
  * Retrieve and manage Alexa ToDo lists and items
  * Full item operations: add, delete, rename, mark complete/incomplete
  * Receive real-time ToDo update notifications
  * Automatic periodic synchronization of ToDo lists

[Review Change Stack](https://app.coderabbit.ai/change-stack/chemelli74/aioamazondevices/pull/839?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


src/aioamazondevices/const/http.py
```diff
@@ -47,6 +47,7 @@
 URI_NEXUS_GRAPHQL = "/nexus/v1/graphql"
 URI_NOTIFICATIONS = "/api/notifications"
 URI_SIGNIN = "/ap/signin"
+URI_LISTS = "/alexashoppinglists/api/v2/lists"
```
chemelli74: Maybe `URI_TODOS` to reflect the usage ? Like we do for `URI_MEDIA_*`

src/aioamazondevices/implementation/todo.py
```diff
@@ -0,0 +1,221 @@
...
+    ) -> None:
+        """Initialize AmazonToDoHandler class."""
+        self._session_state_data = session_state_data
+        self._http_wrapper = http_wrapper
+
+        self._base_url = f"https://www.amazon.{self._session_state_data.domain}"
+
+        self.lists: list[ListInfo] = []
+        self.all_items: dict[str, list[ListItem]] = {}
+        self.all_items_lookup: dict[str, dict[str, ListItem]] = {}
```
chemelli74: Those 3 are needed outside the class for some reason or are only internals as I expect them to be ?

If so, please use the "_" like `self._lists`

src/aioamazondevices/structures.py
```diff
@@ -170,3 +170,72 @@ class AmazonVocalRecord:
...
+@dataclass
+class ListInfo:
+    """Amazon list info."""
+
+    id: str
+    list_type: ListType
+    custom_list_name: str | None
+
+    @property
+    def name(self) -> str:
```
chemelli74: Can we use instead var "name" and make the magic in the code and not in the structure ?

src/aioamazondevices/structures.py
```diff
@@ -170,3 +170,72 @@ class AmazonVocalRecord:
...
+class ListItem:
+    """Amazon list item."""
+
+    id: str
+    status: ListItemStatus
+    original_name: str
+    version: int
+
+    @property
+    def name(self) -> str:
```
chemelli74: Same as above

src/aioamazondevices/structures.py
```diff
@@ -170,3 +170,72 @@ class AmazonVocalRecord:
...
+        """Get the name of the list item with the first letter capitalized.
+
+        Returns:
+            Name of the list item with the first letter capitalized.
+
+        """
+        return self.original_name[0].upper() + self.original_name[1:]
+
+    @property
+    def is_complete(self) -> bool:
```
chemelli74: Not a structure, but a todo.py method, please move it there

src/aioamazondevices/implementation/todo.py
```diff
@@ -0,0 +1,221 @@
...
+        Args:
+            list_id: The ID of the list to fetch items from.
+            limit: The number of items to fetch in each batch.
+                   Defaults to 100 which is the maximum allowed by the API.
+
+        Returns:
+            A list of shopping list items.
+
+        """
+        list_items, _ = await self.get_list_items_check_has_more(list_id, limit)
```
chemelli74: This is a single method call, is that really needed ?

src/aioamazondevices/api.py
```diff
@@ -144,6 +152,21 @@ def routines(self) -> list[str]:
...
+    @property
+    def todo_list_items(self) -> dict[str, list[ListItem]]:
+        """Return all list items."""
+        return self._todo_handler.all_items
+
+    @property
+    def todo_list_items_lookup(self) -> dict[str, dict[str, ListItem]]:
+        """Return all list items lookup."""
+        return self._todo_handler.all_items_lookup
+
```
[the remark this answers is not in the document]
lonlazer: update_todo_lists() is currently not called from outside. Should we still add the public wrapper?

src/aioamazondevices/implementation/todo.py
```diff
@@ -0,0 +1,221 @@
...
+
+        response_json = await self._http_wrapper.response_to_json(raw_resp)
+
+        item_info_list = response_json.get("itemInfoList", [])
+
+        list_items = [
+            ListItem(
+                id=item_info["itemId"],
+                original_name=item_info["itemName"],
+                status=ListItemStatus(item_info["itemStatus"]),
```
chemelli74: Why a class and not directly the value ?

```suggestion
                status=item_info["itemStatus"],
```

src/aioamazondevices/implementation/todo.py
```diff
@@ -0,0 +1,221 @@
...
+
+        response_json = await self._http_wrapper.response_to_json(
+            raw_resp, "listInfoList"
+        )
+        list_info_list = response_json["listInfoList"]
+
+        return [
+            ListInfo(
+                id=list_info["listId"],
+                list_type=ListType(list_info["listType"]),
```
chemelli74: Do we really need a structure for that ? Why not simply:

```suggestion
                list_type=list_info["listType"],
```

src/aioamazondevices/api.py
```diff
@@ -205,6 +229,8 @@ async def get_devices_data(
         # Perform a refresh to ensure your data is as up-to-date as possible.
         await self._refresh_basic_data()
 
+        await self._todo_handler.update_all_items()
```
chemelli74: I think we should use HTTP2 push for refreshing, not polling

chemelli74 (changes requested):
Thank you for your submission !

A few things to check/adjust

src/aioamazondevices/const/http.py
lonlazer: Done. Renamed.

src/aioamazondevices/implementation/todo.py
lonlazer: Done. Removed this function. I was using this before, to print a warning, if a list has more than 100 items. But probably not necessary if we mention it clearly in the documentation.

src/aioamazondevices/implementation/todo.py
lonlazer: @chemelli74 Currently, yes. It is similarly implemented as https://github.com/chemelli74/aioamazondevices/blob/06afe4e346546be59eacf5eb8ffedf2b8f9536fb/src/aioamazondevices/implementation/sequence.py#L53-L55, where the routines property is directly accessed by the integration.

Should this be stored elsewhere?

src/aioamazondevices/implementation/todo.py
lonlazer: @chemelli74 It can only have the values ACTIVE and COMPLETE, so I thought it would be cleaner to have it as Enum that the value can be easily compared.

Do you want me to change it to str and compare the strings elsewhere?

src/aioamazondevices/implementation/todo.py
```diff
@@ -0,0 +1,214 @@
...
+        )
+
+        response_json = await self._http_wrapper.response_to_json(raw_resp)
+
+        item_info_list = response_json.get("itemInfoList", [])
+
+        return [
+            ListItem(
+                id=item_info["itemId"],
+                name=(item_info["itemName"][0].upper() + item_info["itemName"][1:])
```
chemelli74: Can you elaborate why we need to manipulate the result ?

lonlazer: When items are added by Voice, they are always lowercase. IMHO that doesn't look nice in the UI.

I put the code in a separate function to make in cleaner.

But of course, it is not 100 % necessary and could be removed again

src/aioamazondevices/api.py
lonlazer: Implemented :)

src/aioamazondevices/implementation/todo.py
chemelli74: You can then simply use
```python
item_info["itemName"].capitalize()
```

src/aioamazondevices/implementation/todo.py
chemelli74: We can use const for that, I would say a strEnum for 2 values in overstructured

src/aioamazondevices/implementation/todo.py
chemelli74: So make them properties

src/aioamazondevices/api.py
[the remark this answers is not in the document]
chemelli74: Everything that is only library internal should have a underscore prefix.

src/aioamazondevices/implementation/todo.py
lonlazer: Yes, that would work, but it is not optimal for languages with noun capitalization such as German.

`capitalize()` would turn

"Die Katze streicheln"

into

"Die katze streicheln".

But I agree that this in an edge case. We could replace it with capitalize()

src/aioamazondevices/api.py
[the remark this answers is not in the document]
lonlazer: Done

src/aioamazondevices/implementation/todo.py
lonlazer: Done

src/aioamazondevices/implementation/todo.py
lonlazer: Done

src/aioamazondevices/implementation/todo.py
lonlazer: Done

src/aioamazondevices/implementation/todo.py
```diff
@@ -0,0 +1,236 @@
...
+    Returns:
+        True if the list item is complete, False otherwise.
+
+    """
+    return list_item.status == LIST_ITEM_STATUS_COMPLETE
+
+
+def _capitalize_first_letter(text: str) -> str:
+    """Capitalize the first letter of a string."""
+    return (text[0].upper() + text[1:]) if text else ""
```
chemelli74: Can we use `capitalize()` and avoid the method completely ?

src/aioamazondevices/implementation/todo.py
```diff
@@ -0,0 +1,236 @@
...
+            A list of shopping list information objects.
+
+        Raises:
+            Exception: If the API request fails.
+
+        """
+        _, raw_resp = await self._http_wrapper.session_request(
+            HTTPMethod.POST,
+            f"{self._base_url}{URI_TODO}/fetch",
+            input_data={},
```
chemelli74: Do we need to specify a empty dict here ?

src/aioamazondevices/implementation/todo.py
```diff
@@ -0,0 +1,236 @@
...
+        )
+        list_info_list = response_json["listInfoList"]
+
+        return [
+            ListInfo(
+                id=list_info["listId"],
+                list_type=list_info["listType"],
+                name=list_info["listName"]
+                if list_info["listType"] == LIST_TYPE_CUSTOM
+                else list_info["listType"].capitalize(),
```
chemelli74: Not a fun of such manipulation, can we avoid it ?

src/aioamazondevices/implementation/todo.py
```diff
@@ -0,0 +1,236 @@
...
+                   Defaults to 100 which is the maximum allowed by the API.
+
+        Returns:
+            A list of list items.
+
+        """
+        _, raw_resp = await self._http_wrapper.session_request(
+            HTTPMethod.POST,
+            f"{self._base_url}{URI_TODO}/{list_id}/items/fetch?limit={limit}",
+            input_data={},
```
chemelli74: Same as above

src/aioamazondevices/implementation/todo.py
```diff
@@ -0,0 +1,236 @@
...
+        )
+
+        response_json = await self._http_wrapper.response_to_json(raw_resp)
+
+        item_info_list = response_json.get("itemInfoList", [])
+
+        return [
+            ListItem(
+                id=item_info["itemId"],
+                name=_capitalize_first_letter(item_info["itemName"]),
```
chemelli74:
```suggestion
                name=item_info["itemName"].capitalize(),
```

src/aioamazondevices/implementation/todo.py
```diff
@@ -0,0 +1,236 @@
...
+
+        """
+        await self._http_wrapper.session_request(
+            HTTPMethod.POST,
+            f"{self._base_url}{URI_TODO}/{list_id}/items",
+            input_data={
+                "items": [
+                    {
+                        "itemType": "KEYWORD",
+                        "itemName": name,
```
chemelli74, line 163: Is the same name we capitalized before ? Isn't that inconsistent ?

src/aioamazondevices/implementation/todo.py
```diff
@@ -0,0 +1,236 @@
...
+            item_id: The ID of the item to delete.
+            version: The current version of the item.
+                     The value is included in the get_list_items response and is
+                     required by the Amazon API.
+
+        """
+        await self._http_wrapper.session_request(
+            HTTPMethod.DELETE,
+            f"{self._base_url}{URI_TODO}/{list_id}/items/{item_id}?version={version}",
+            input_data={},
```
chemelli74: Samve as above

src/aioamazondevices/implementation/todo.py
```diff
@@ -0,0 +1,236 @@
...
+            input_data={
+                "itemAttributesToUpdate": [{"type": "itemName", "value": new_name}],
+                "itemAttributesToRemove": [],
+            },
+            json_data=True,
+        )
+
+    async def update_lists(self) -> None:
+        """Update the lists."""
+        self._lists = await self.get_lists()
```
chemelli74: Why we need such method ? Cannot we call directly get_lists() and make it update internals ?

src/aioamazondevices/implementation/todo.py
```diff
@@ -0,0 +1,236 @@
...
+
+    async def sync_all_items(self, list_id: str | None = None) -> None:
```
chemelli74: Cannot we remove this method and instead call get_list_items only when we have a list_id ?

src/aioamazondevices/api.py
```diff
@@ -281,6 +301,11 @@ async def _http2_push_event_handler(
             await self._emit_media_state_event()
             await self._emit_history_event()
             return
+        if event_type == AmazonPushMessage.ItemChange.value:
+            list_id = payload.get("listId")
+            await self._todo_handler.sync_all_items(list_id=list_id)
```
chemelli74: Does the event only inform of a list update, but not which item ?

src/aioamazondevices/api.py
```diff
@@ -447,6 +472,19 @@ async def _emit_volume_state_event(self) -> None:
...
+    async def _emit_todo_event(self, list_id: str | None = None) -> None:
+        """Emit todo event to subscribers."""
+        if self.on_todo_event.frozen:
+            _LOGGER.debug("Emitting todo event to subscribers")
+
+            if list_id:
+                # Only send the list which was actually updated
+                all_items = {list_id: self._todo_handler.all_items.get(list_id, [])}
+            else:
+                all_items = self._todo_handler.all_items
```
chemelli74: Seems to me that all_items should handle such logic

src/aioamazondevices/implementation/todo.py
```diff
@@ -0,0 +1,236 @@
...
+                        if checked
+                        else LIST_ITEM_STATUS_ACTIVE,
+                    }
+                ],
+                "itemAttributesToRemove": [],
+            },
+            json_data=True,
+        )
+
+    async def add_item(self, list_id: str, name: str) -> None:
```
chemelli74, line 148: CAn we consolidate those methods with a param avoiding repeating similar code  all over ?

src/aioamazondevices/const/todo.py
```diff
@@ -0,0 +1,8 @@
+"""Todo constants."""
+
+LIST_ITEM_STATUS_ACTIVE = "ACTIVE"
+LIST_ITEM_STATUS_COMPLETE = "COMPLETE"
+
+LIST_TYPE_CUSTOM = "CUSTOM"
+LIST_TYPE_SHOP = "SHOP"
+LIST_TYPE_TODO = "TODO"
```
jamesonuk: Would this be better as an enum as presumably it is used to determine list type and a list can only have one type?

src/aioamazondevices/structures.py
```diff
@@ -170,3 +170,22 @@ class AmazonVocalRecord:
...
+    list_type: Literal["SHOP", "CUSTOM", "TODO"]
+    name: str
+
+
+@dataclass
+class ListItem:
+    """Amazon list item."""
+
+    id: str
+    status: Literal["ACTIVE", "COMPLETE"]
```
jamesonuk: Same as in const.   Is this better as an enum

src/aioamazondevices/implementation/todo.py
lonlazer: Yes, otherwise this will happen:

```
Traceback (most recent call last):
  File "/home/leon/Repos/aioamazondevices/test_todo_stuff.py", line 138, in <module>
    asyncio.run(main())
    ~~~~~~~~~~~^^^^^^^^
  File "/home/leon/.local/share/uv/python/cpython-3.14.2-linux-x86_64-gnu/lib/python3.14/asyncio/runners.py", line 204, in run
    return runner.run(main)
           ~~~~~~~~~~^^^^^^
  File "/home/leon/.local/share/uv/python/cpython-3.14.2-linux-x86_64-gnu/lib/python3.14/asyncio/runners.py", line 127, in run
    return self._loop.run_until_complete(task)
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
  File "/home/leon/.local/share/uv/python/cpython-3.14.2-linux-x86_64-gnu/lib/python3.14/asyncio/base_events.py", line 719, in run_until_complete
    return future.result()
           ~~~~~~~~~~~~~^^
  File "/home/leon/Repos/aioamazondevices/test_todo_stuff.py", line 83, in main
    await api._todo_handler.update_lists()
  File "/home/leon/Repos/aioamazondevices/src/aioamazondevices/implementation/todo.py", line 223, in update_lists
    self._lists = await self.get_lists()
                  ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/leon/Repos/aioamazondevices/src/aioamazondevices/implementation/todo.py", line 70, in get_lists
    _, raw_resp = await self._http_wrapper.session_request(
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<4 lines>...
    )
    ^
  File "/home/leon/Repos/aioamazondevices/src/aioamazondevices/http_wrapper.py", line 383, in session_request
    raise CannotRetrieveData(
        f"Request failed: {await self.http_phrase_error(resp.status)}"
    )
src.aioamazondevices.exceptions.CannotRetrieveData: Request failed: Bad Request
```

lonlazer: Probably because of:
https://github.com/lonlazer/aioamazondevices/blob/996ce188e90528eb5b175e3c4604a99309db6340/src/aioamazondevices/http_wrapper.py#L339

```
>>> orjson.dumps(None)
b'null'
>>> orjson.dumps({})
b'{}'
>>> 
```
null will not be accepted by the API

src/aioamazondevices/implementation/todo.py
lonlazer: See: https://github.com/chemelli74/aioamazondevices/pull/839#discussion_r3283088569

You reacted with thumbs up, I thought you accepted this reason.

Or do you think we should ignore this edge case?

src/aioamazondevices/implementation/todo.py
lonlazer: See https://github.com/chemelli74/aioamazondevices/pull/839#discussion_r3283899591

src/aioamazondevices/implementation/todo.py
lonlazer: What would you prefer? This will be the list name displayed to the user.

If it is not a custom list, the name field will be returned empty by the API. So too fill it, i will just use the type.

Maybe we leave the name empty for non-custom list and handle this in the integration (maybe using string.json)?

src/aioamazondevices/implementation/todo.py
lonlazer: The capitalization is done to show it capitalized in the frontend, so only on the way Alexa -> Hass

If items are added by voice, they will be lower case.

This method here is for Hass -> Alexa, here not capitalization is done.

src/aioamazondevices/structures.py
lonlazer: Actually I agree. I used an enum before. 
But was asked to remove it: https://github.com/chemelli74/aioamazondevices/pull/839#discussion_r3261712241

lonlazer: Hmm hass also uses enum for a similar use case: https://github.com/home-assistant/core/blob/49fc1b413d0446092cdb8575643fe81e53c844df/homeassistant/components/todo/const.py#L47

src/aioamazondevices/api.py
lonlazer: The event actually includes the updated item id, but not the changed attributes (new name or status...).

Furthermore, even if we could just retrieve the updated item, it would be not so useful as Home Assistant stores the items in a list, and not in a dict. So replacing the changed item is not feasible (would require iterating and comparing the IDs, as the order of the list could have changed)

src/aioamazondevices/implementation/todo.py
lonlazer: Done

src/aioamazondevices/api.py
lonlazer: Hmm, not sure if I understand this correctly. all_items always contains all items.

Only the todo_event has two cases
- Only one list  updated (by adding etc.) 
- Update all lists (only done for initial sync)

src/aioamazondevices/implementation/todo.py
lonlazer: Like this?

 ```python
async def _call_todo_api(self, url, input_data):
     _, raw_resp = await self._http_wrapper.session_request(....
    response_json = await self._http_wrapper.response_to_json(raw_resp)
    return response_json

async def set_item_checked_status(
        self, list_id: str, item_id: str, checked: bool, version: int
    ) -> None:
        """Update the checked status of an item in a shopping list.

        Args:
            list_id: The ID of the list containing the item.
            item_id: The ID of the item to update.
            checked: True to mark as complete, False to mark as active.
            version: The current version of the item.
                     The value is included in the get_list_items response and is
                     required by the Amazon API.

        """
        await self._call_todo_api(
            f"/{list_id}/items/{item_id}?version={version}",
            input_data={
                "itemAttributesToUpdate": [
                    {
                        "type": "itemStatus",
                        "value": LIST_ITEM_STATUS_COMPLETE
                        if checked
                        else LIST_ITEM_STATUS_ACTIVE,
                    }
                ],
                "itemAttributesToRemove": [],
            },
        )

async def add_item(...

...
```

+

File: src/aioamazondevices/structures.py (hunk 25 above)
Comment (lonlazer) [reply src/aioamazondevices/structures.py]: @chemelli74 What is your viewpoint here? If [Home Assitant core](https://github.com/home-assistant/core/blob/49fc1b413d0446092cdb8575643fe81e53c844df/homeassistant/components/todo/const.py#L47) also uses StrEnum for the exact same use case, can we also use it?

File: src/aioamazondevices/implementation/todo.py (hunk 23 above)
Comment (lonlazer) [reply src/aioamazondevices/implementation/todo.py:148]: Hi @jamesonuk what is your opinion on this? We discussed this already on Discord a bit.

The proposal of putting everything in one update method would result into: 


```
It needs  parameters for the new name and for new checked status

update(method, item_id, new_state=False, new_name="")


Call

update(method=RENAME, new_name="Milk")
update(method=CHECK, new_state=True)

I feel separate methods are easier to understand.
```

File: src/aioamazondevices/implementation/todo.py (hunk 14 above)
Comment (chemelli74) [reply src/aioamazondevices/implementation/todo.py]: Yup, was my idea. Let's code it and see how it looks like (user and code wise) before changing code here

File: src/aioamazondevices/implementation/todo.py (hunk 13 above)
Comment (chemelli74) [reply src/aioamazondevices/implementation/todo.py]: Thx for the double check. I would just add a comment there for future review

File: src/aioamazondevices/implementation/todo.py (hunk 16 above)
Comment (chemelli74) [reply src/aioamazondevices/implementation/todo.py]: We should do it the same all over. I would say capitalize() but eventually we decide for the custom method, use it in all calls

File: src/aioamazondevices/implementation/todo.py (hunk 17 above)
Comment (chemelli74) [reply src/aioamazondevices/implementation/todo.py:163]: I mean ifwe changed the capitalization in the first run ( library -> HomeAssistant ) when we do it for the toehr way we found it capitalized and Alexa expect it lower case ?

File: src/aioamazondevices/api.py (hunk 21 above)
Comment (chemelli74) [reply src/aioamazondevices/api.py]: But we can store the data in libray and in internal HomeAssistant class as dict so to allow easiy update, then we can convert dict to list for the public HomeAssistant method call

File: src/aioamazondevices/api.py (hunk 22 above)
Comment (chemelli74) [reply src/aioamazondevices/api.py]: make all items a method and return data based on a parameter, then remove completely this if

File: src/aioamazondevices/structures.py (hunk 25 above)
Comment (chemelli74) [reply src/aioamazondevices/structures.py]: You have 2 votes, so go for the Enum. Sorry for the extra work ;-)

File: src/aioamazondevices/implementation/todo.py [hunk 26]
```diff
@@ -0,0 +1,224 @@
...
+    """Check if a list item is complete.
+
+    Args:
+        list_item: The list item to check.
+
+    Returns:
+        True if the list item is complete, False otherwise.
+
+    """
+    return list_item.status == ListItemStatus.COMPLETE
```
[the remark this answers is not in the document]
lonlazer: Can you commit this?

src/aioamazondevices/implementation/todo.py
lonlazer: The Alexa API will accept anything. You can create uppercase items via the Alexa app.

This is only for the case:
- Item created by voice via Alexa (will be lowercase)
- Show in uppercase in Home Assistant frontend for aesthetic reasons

If the item is created on the Home Assitant side, no changes have to be made on the way back to the Alexa API as the API accepps anything.

Right now user will have already mixed lists: Lowercase by creating by voice, and possibly uppercase if they use the Alexa app

src/aioamazondevices/implementation/todo.py
lonlazer: For the initial sync we need to sync all available list, so not list_id is given;

https://github.com/lonlazer/alexa_devices_test/blob/6106ddf445d9837ca6bb525915474c2d894205e8/custom_components/alexa_devices/__init__.py#L48

(similar to the media state sync)

Should I refactor this somehow? Like to separate methods or anything?

src/aioamazondevices/implementation/todo.py
lonlazer: Done, added comment

src/aioamazondevices/structures.py
lonlazer: No problem :)
Done, reverted back to Enum

src/aioamazondevices/implementation/todo.py
lonlazer: Okay, we could remove that logic from the libray:

```python
        self._lists = [
            ListInfo(
                id=list_info["listId"],
                list_type=list_info["listType"],
                name=list_info.get("listName", None)
            )
            for list_info in list_info_list
        ]
```

And then do in the integration something like this:
```python
        if alexa_list.list_type == ListType.SHOP:
            entity_description = EntityDescription(key=alexa_list.id,
                                                   translation_key="shop")

        elif alexa_list.list_type == ListType.TODO:
            entity_description = EntityDescription(key=alexa_list.id,
                                                   translation_key="todo")
        else:
            # Custom list -> Use actual name
            entity_description = EntityDescription(key=alexa_list.id,
                                                   name=alexa_list.name)
```

How does that sound?

chemelli74: Sounds nice to me!

src/aioamazondevices/implementation/todo.py
lonlazer: Other occurence of capitalize() removed (list naming logic now in integration)

src/aioamazondevices/api.py
lonlazer: Done: Moved the If from _emit to get_todo_list_items (formerly all_items)

src/aioamazondevices/implementation/todo.py
lonlazer: Nvm, removed this function now all together

src/aioamazondevices/api.py
lonlazer: Okay, I implemented it like this:

refactor(todo): implement push mechanism for single items
- Removed item sync functions and all_items
- Centralized the storage of full lists within the integration's coordinator.py.
- Updated the event listener to fetch the list upon receiving an event, ensuring only the specific changed item is pushed back to Home Assistant.

src/aioamazondevices/implementation/todo.py
```diff
@@ -24,7 +24,10 @@ def is_item_complete(list_item: ListItem) -> bool:
 
 
 def _capitalize_first_letter(text: str) -> str:
-    """Capitalize the first letter of a string."""
+    """Capitalize the first letter of a string.
+
+    In contrast to capitalize(), this function keeps the remaining letters untouched.
```
chemelli74: IF we do for a proper UI display, I would go with capitalize() honestly

src/aioamazondevices/api.py
```diff
@@ -297,9 +299,24 @@ async def _http2_push_event_handler(
...
+
+            _LOGGER.info("Received ItemChange for %s: %s", list_id, list_event_type)
+
+            if list_event_type == ListEventType.DELETED:
+                list_event = ListEvent(list_id, item_id, list_event_type)
+            elif list_event_type in (ListEventType.UPDATED, ListEventType.CREATED):
+                list_items = await self.get_todo_list_items(list_id)
+
+                list_event = ListEvent(
+                    list_id, item_id, list_event_type, item=list_items[item_id]
```
chemelli74: Maybe` items=` ?

src/aioamazondevices/api.py
```diff
@@ -467,13 +484,11 @@ async def _emit_volume_state_event(self) -> None:
...
 
-    async def _emit_todo_event(self, list_id: str | None = None) -> None:
+    async def _emit_todo_event(self, list_event: ListEvent) -> None:
         """Emit todo event to subscribers."""
         if self.on_todo_event.frozen:
-            _LOGGER.debug("Emitting todo event to subscribers")
-            items = self.get_todo_list_items(list_id=list_id)
-
-            await self.on_todo_event.send(items)
+            _LOGGER.debug("Emitting todo event: %s for %s", list_event)
```
chemelli74: You have 2 placeholders and only 1 var

src/aioamazondevices/implementation/todo.py
```diff
@@ -0,0 +1,221 @@
...
+from typing import Any
+
+from aiohttp import ClientResponse
+
+from aioamazondevices.const.http import URI_TODO
+from aioamazondevices.http_wrapper import AmazonHttpWrapper, AmazonSessionStateData
+from aioamazondevices.structures import ListInfo, ListItem, ListItemStatus, ListType
+
+
+def is_item_complete(list_item: ListItem) -> bool:
```
chemelli74: Seems not used in the library

src/aioamazondevices/implementation/todo.py
```diff
@@ -0,0 +1,221 @@
...
+
+        """
+        if input_data is None:
+            input_data = {}  # Required by API, otherwise call will be rejected
+        _, raw_response = await self._http_wrapper.session_request(
+            method,
+            f"{self._base_url}{URI_TODO}/{url}",
+            input_data=input_data,
+            json_data=True,
+        )
```
chemelli74:
```suggestion
        _, raw_response = await self._http_wrapper.session_request(
            method,
            f"{self._base_url}{URI_TODO}/{url}",
            input_data=input_data or {}, # API doesn't allone 'None'
            json_data=True,
        )
```

src/aioamazondevices/api.py
chemelli74: I like the implementation very much. Thx !

src/aioamazondevices/implementation/todo.py
chemelli74: @jamesonuk your call here

chemelli74:
A few small things and I would say we are good to go.

Can I has you to swap `method `and `url` params for `_call_lists_api` to be consistent:

https://github.com/chemelli74/aioamazondevices/blob/0a0dcd2687ea4236640767537099cebc779610b6/src/aioamazondevices/http_wrapper.py#L281-L288

src/aioamazondevices/api.py
```diff
@@ -281,6 +298,26 @@ async def _http2_push_event_handler(
...
+
+            _LOGGER.info("Received ItemChange for %s: %s", list_id, list_event_type)
+
+            if list_event_type == ListEventType.DELETED:
+                list_event = ListEvent(list_id, item_id, list_event_type)
+            elif list_event_type in (ListEventType.UPDATED, ListEventType.CREATED):
+                list_items = await self.get_todo_list_items(list_id)
+
+                list_event = ListEvent(
+                    list_id, item_id, list_event_type, item=list_items[item_id]
```
```suggestion
            if list_event_type not in ListEventType:
                _LOGGER.warning(
                    "Received unsupported list event type: %s", list_event_type
                )
                return

            items = None
            if list_event_type != ListEventType.DELETED:
                list_items = await self.get_todo_list_items(list_id)
                items = list_items[item_id]

            list_event = ListEvent(list_id, item_id, list_event_type, items=items)
```

src/aioamazondevices/implementation/todo.py
jamesonuk: Can't really dig in depth on my phone but unless there is any technical reason to do so (like what gets displayed in the app is formatted differently to what is returned by the API, uniqueness etc.) I would not touch it at all.

HA and Alexa app should really show the same.

lonlazer: Okay, two votes. Then I will remove this

src/aioamazondevices/implementation/todo.py
lonlazer: It is just called by the integration. I could also just compare the enum values there. As this is needed twice, I thought it's cleaner in a function to avoid code duplicates.

chemelli74: I think such method could be a single line check in HomeAssistant side

chemelli74 (approved):
Nice work !

PR #858, opened 2026-06-01
fix!: handle recordType history records

jamesonuk:
## Summary by CodeRabbit

* **Bug Fixes**
  * Vocal history parsing is more resilient: malformed or missing fields (including absent device info) are safely skipped to prevent errors and preserve per-device behavior.

* **New Features**
  * Records now expose a single consolidated record-type field and use safer fallbacks for intent, timestamp, title and subtitle, improving accuracy and stability of recent voice-interaction history.

### DIFF
--- a/src/aioamazondevices/implementation/history.py
+++ b/src/aioamazondevices/implementation/history.py
@@ -77,15 +77,15 @@ class AmazonHistoryHandler:
         records: dict[str, AmazonVocalRecord] = {}
         for record in history_json["alexaHistoryRecords"]:
             _LOGGER.debug("Processing vocal history record: %s", record)
-            utterance_type = record["utteranceType"]
+            utterance_type = record.get("utteranceType")
             if (
                 utterance_type
-                in [
+                in (
                     "ASR_TIMEOUT",
                     "DEVICE_ARBITRATION",
                     "NO_EXPRESSED_INTENT",
                     "WAKE_WORD_ONLY",
-                ]
+                )
                 # InvokeRoutineIntent, AddToListIntent are not linked to a device
                 or record["deviceInfo"] is None
             ):
@@ -95,10 +95,11 @@ class AmazonHistoryHandler:
             timestamp = record["timestamp"]
             new_record = AmazonVocalRecord(
                 timestamp=timestamp,
+                record_type=record.get("recordType", "UNKNOWN"),
                 utterance_type=utterance_type,
-                intent=record["intent"],
-                title=record["title"],
-                sub_title=record["subTitle"],
+                intent=record.get("intent"),
+                title=record.get("title"),
+                sub_title=record.get("subTitle"),
             )
             # Store only the latest record per serial number
             if serial not in records or timestamp > records[serial].timestamp:
--- a/src/aioamazondevices/implementation/history.py
+++ b/src/aioamazondevices/implementation/history.py
@@ -100,9 +100,7 @@ class AmazonHistoryHandler:
             timestamp = record["timestamp"]
             new_record = AmazonVocalRecord(
                 timestamp=timestamp,
-                history_type=record.get("utteranceType")
-                or record.get("recordType")
-                or "Unknown",
+                history_type=utterance_type or record.get("recordType") or "Unknown",
                 intent=record.get("intent") or "Unknown",
                 title=record["title"],
                 sub_title=record["subTitle"],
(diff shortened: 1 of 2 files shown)


src/aioamazondevices/implementation/history.py
```diff
@@ -77,7 +77,7 @@ async def get_vocal_history(self) -> dict[str, AmazonVocalRecord]:
         records: dict[str, AmazonVocalRecord] = {}
         for record in history_json["alexaHistoryRecords"]:
             _LOGGER.debug("Processing vocal history record: %s", record)
-            utterance_type = record["utteranceType"]
+            utterance_type = record.get("utteranceType")
```
chemelli74, line 80:
```suggestion
            voice_type = record.get("utteranceType") or record.get("recordType")
```

src/aioamazondevices/implementation/history.py
```diff
@@ -87,18 +87,19 @@ async def get_vocal_history(self) -> dict[str, AmazonVocalRecord]:
                     "WAKE_WORD_ONLY",
                 ]
                 # InvokeRoutineIntent, AddToListIntent are not linked to a device
-                or record["deviceInfo"] is None
+                or record.get("deviceInfo") is None
```
chemelli74: Mot needed, DeviceInfo is always there

src/aioamazondevices/implementation/history.py
```diff
@@ -87,18 +87,19 @@ async def get_vocal_history(self) -> dict[str, AmazonVocalRecord]:
...
             serial = record["deviceInfo"]["deviceSerialNumber"]
             timestamp = record["timestamp"]
             new_record = AmazonVocalRecord(
                 timestamp=timestamp,
+                record_type=record.get("recordType", "UNKNOWN"),
                 utterance_type=utterance_type,
-                intent=record["intent"],
-                title=record["title"],
-                sub_title=record["subTitle"],
+                intent=record.get("intent"),
```
chemelli74:
```suggestion
                intent=record.get("intent") or record.get("type"),
```

src/aioamazondevices/implementation/history.py
```diff
@@ -87,18 +87,19 @@ async def get_vocal_history(self) -> dict[str, AmazonVocalRecord]:
...
+                title=record.get("title"),
+                sub_title=record.get("subTitle"),
```
chemelli74: Not needed, those are the voice command and it's reply. Are always there even if reply may be null but not missing

src/aioamazondevices/structures.py
```diff
@@ -166,10 +166,11 @@ class AmazonVocalRecord:
...
     timestamp: int
-    utterance_type: str
-    intent: str
-    title: str
-    sub_title: str
+    record_type: str
+    utterance_type: str | None
+    intent: str | None
+    title: str | None
+    sub_title: str | None
```
chemelli74: Don't want to be None when not necessary.
Will make code and typing very bad, especially on HomeAssistant side

chemelli74 (changes requested):
As per my idea in Discord, but please review the details

src/aioamazondevices/implementation/history.py
jamesonuk: but does this make a difference.  the logic here is to exclude anything where we don't have `deviceInfo` data.
This is safer as it excluded both cases where `deviceInfo` is there but empty and where it is missing completely.
If a new record type appears this can break the whole integration with a `KeyError`

jamesonuk:
Fixes https://github.com/home-assistant/core/issues/172767

Anonghost720:
**Still crashes after #858 — `deviceInfo` can be a list**

After #858, setup still fails:

```
TypeError: list indices must be integers or slices, not str
  history.py, line 94, in get_vocal_history
```

For `recordType: "conversation"` records, `deviceInfo` is returned as a **list**, not a dict:

```json
"deviceInfo": [{"deviceSerialNumber": "...PRDSPK1", "deviceType": "...", "deviceName": "..."}]
```

But the code does `record["deviceInfo"]["deviceSerialNumber"]`, which assumes a dict → `TypeError`.

**Fix:** normalize before indexing:

```python
device_info = record.get("deviceInfo")
if isinstance(device_info, list):
    device_info = device_info[0] if device_info else None
if not isinstance(device_info, dict):
    continue
serial = device_info["deviceSerialNumber"]
```

Confirmed working locally on HA Core `2026.6.0b2` with this patch.

chemelli74 (approved):
I'm not 100% happy with the code.

But approving anyway as I have no PC in my hands to refine and new release of HomeAssistant will be tomorrow.

PR #869, opened 2026-06-04
fix: move device capabilities registration during login

chemelli74:
## Summary by CodeRabbit

* **Improvements**
  * Device capability registration is now integrated into the login process for better flow.
  * Simplified HTTP/2 connection handling to improve stability and reliability.
  * Enhanced token refresh with improved error handling and logging.


src/aioamazondevices/login.py
```diff
@@ -261,6 +281,8 @@ async def login_mode_interactive(self, otp_code: str) -> dict[str, Any]:
         login_data = await self._register_device(device_login_data)
         self._session_state_data.login_stored_data = login_data
 
+        await self._register_device_capabilities()
```
jamesonuk: I would possibly put this after the cookies have been refreshed.    Probably makes little difference but probably makes sense to refresh cookies before this.

src/aioamazondevices/implementation/http2.py
```diff
@@ -315,29 +287,16 @@ async def _refresh_token(self) -> bool:
...
-                _LOGGER.warning("Failed to register device capabilities: %s", exc)
-                return False
-
-        return True
-
     async def _stream_and_process(self) -> None:
         """Open stream and process incoming directives."""
         _LOGGER.debug("Starting AVS Directives stream")
 
+        bearer = f"Bearer {http2_get_bearer_token(self._session_state_data)}"
```
jamesonuk: Same code on line 428.    See utils for possible better pattern

src/aioamazondevices/utils.py
```diff
@@ -256,3 +258,11 @@ def _replace(
         return value
 
     return {k: _replace(v) for k, v in obj.items()}
+
+
+def http2_get_bearer_token(session_state_data: AmazonSessionStateData) -> str:
```
jamesonuk: This isn't http2.    This token is potentially used across different restful calls too.

src/aioamazondevices/utils.py
```diff
@@ -256,3 +258,11 @@ def _replace(
...
+    """Return the current access token."""
```
jamesonuk: Wondering whether this is actually better in `AmazonSessionStateData` as a property?
Call then becomes something like
```python
headers={"Authorization": f"Bearer {self._session_state_data.bearer_token}"},
```
rather than passing around`_session_state_data`
```python
bearer = f"Bearer {http2_get_bearer_token(self._session_state_data)}"
"Authorization": f"{bearer}",
```
or even possibly cleaner to return the full header in property so call would be 

```python
headers={"Authorization": self._session_state_data.bearer_token,
```

src/aioamazondevices/utils.py
chemelli74: We use it only for that so far ;-)

jamesonuk (approved):

PR #874, opened 2026-06-08
fix: use the correct AVS site

chemelli74:
## Summary by CodeRabbit

* **Refactor**
  * Updated endpoint configuration for improved service connectivity.
  * Enhanced dynamic endpoint detection during streaming sessions to maintain connection reliability and automatically correct connectivity issues without manual intervention.

* **Bug Fixes**
  * Improved keepalive ping handling to use the correct, dynamically-detected service endpoint.

### DIFF
--- a/src/aioamazondevices/implementation/http2.py
+++ b/src/aioamazondevices/implementation/http2.py
@@ -305,12 +305,28 @@ class AmazonHTTP2Client:
 
         _LOGGER.debug("Attempting to extract AVS site from response body")
 
-        json_extracted = response.content[
-            response.content.find(b"{") : response.content.rfind(b"}") + 1
-        ]
-        _LOGGER.debug("Extracted JSON from response: %s", json_extracted)
+        boundary = http2_parse_boundary_delimiter(
+            response.headers.get("content-type", "")
+        )
+        if boundary is None:
+            _LOGGER.warning("Missing multipart boundary in SynchronizeState response")
+            return
 
-        site = orjson.loads(json_extracted)["directive"]["payload"]["endpoint"]
+        parser = AvsDirectiveStreamParser(boundary)
+        parts = parser.feed(response.content)
+        chunk_json = next(
+            (http2_extract_json_from_part(p) for p in parts if p),
+            None,
+        )
+        if chunk_json is None:
+            _LOGGER.warning("Could not extract JSON from SynchronizeState response")
+            return
+
+        site = chunk_json.get("directive", {}).get("payload", {}).get("endpoint")
+        if not site:
+            _LOGGER.warning("Could not extract AVS site from SynchronizeState response")
+            return
+
         self._avs_directive_site = site
         _LOGGER.debug("Updated AVS directive site: %s", site)
         raise UpdatedAVSSite
--- a/src/aioamazondevices/implementation/http2.py
+++ b/src/aioamazondevices/implementation/http2.py
@@ -272,7 +272,6 @@ class AmazonHTTP2Client:
         the correct site from the response and update it for future requests.
         """
         url = f"{self._avs_directive_site}/v{HTTP2_DIRECTIVES_VERSION}/events"
-        boundary = str(uuid.uuid4())
 
         metadata = {
             "context": [],
@@ -286,27 +285,14 @@ class AmazonHTTP2Client:
             },
         }
 
-        metadata_json = orjson.dumps(metadata)
-
-        data = b"".join(
-            [
-                f"--{boundary}\r\n".encode(),
-                b'Content-Disposition: form-data; name="metadata"\r\n',
-                f"Content-Length: {len(metadata_json)}\r\n".encode(),
-                b"\r\n",
-                metadata_json,
-                b"\r\n",
-                f"--{boundary}--\r\n".encode(),
-            ]
-        )
+        files = {
+            "metadata": (None, orjson.dumps(metadata), "application/json"),
+        }
 
         response = await self._http2_client.post(
             url,
-            headers={
-                "Authorization": f"Bearer {self._get_bearer_token()}",
-                "Content-Type": f"multipart/form-data; boundary={boundary}",
-            },
-            content=data,
+            headers={"Authorization": f"Bearer {self._get_bearer_token()}"},
+            files=files,
             timeout=httpx.Timeout(30),
         )
         _LOGGER.debug(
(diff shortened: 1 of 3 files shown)


src/aioamazondevices/implementation/http2.py
```diff
@@ -255,6 +264,65 @@ async def _run_tasks(self) -> None:
...
+                b'Content-Disposition: form-data; name="metadata"\r\n',
+                f"Content-Length: {len(metadata_json)}\r\n".encode(),
+                b"\r\n",
+                metadata_json,
+                b"\r\n",
+                f"--{boundary}--\r\n".encode(),
+            ]
+        )
+
+        response = await self._http2_client.post(
```
jamesonuk, line 292: We should set a timeout here as the client has been created with an infinite timeout for the directives loo.
We can and should override this for this request

`timeout=httpx.Timeout(30),` (or probably less, not sure what the default is)

src/aioamazondevices/implementation/http2.py
```diff
@@ -255,6 +264,65 @@ async def _run_tasks(self) -> None:
...
+
+        _LOGGER.debug("Attempting to extract AVS site from response body")
+
+        json_extracted = response.content[
+            response.content.find(b"{") : response.content.rfind(b"}") + 1
+        ]
+        _LOGGER.debug("Extracted JSON from response: %s", json_extracted)
+
+        site = orjson.loads(json_extracted)["directive"]["payload"]["endpoint"]
+        self._avs_directive_site = site
```
jamesonuk: We should be able to reuse the existing functions rather than fragile find.

```python
chunk_json = http2_extract_json_from_part(response.content)
if chunk_json is None:
    _LOGGER.warning("Could not extract JSON from SynchronizeState response")
    # need error here ?? 
    return

site = chunk_json.get("directive", {}).get("payload", {}).get("endpoint")
# check we actually have site? or use ["directive"]["payload"]["endpoint"] and let TypeError/KeyError propagate
self._avs_directive_site = site
```

src/aioamazondevices/implementation/http2.py
```diff
@@ -255,6 +264,65 @@ async def _run_tasks(self) -> None:
                 await asyncio.sleep(delay)
                 self._reconnect_attempt += 1
 
+    async def _get_avs_site(self) -> None:
```
jamesonuk: Probably should be renamed.  It is no longer getting the site.
It is checking whether we need to change AVS site.
perhaps `_check_avs_site` with a docstring explaining as it will definitely be something we forget down the line....

src/aioamazondevices/implementation/http2.py
```diff
@@ -255,6 +264,65 @@ async def _run_tasks(self) -> None:
...
+            [
+                f"--{boundary}\r\n".encode(),
+                b'Content-Disposition: form-data; name="metadata"\r\n',
+                f"Content-Length: {len(metadata_json)}\r\n".encode(),
+                b"\r\n",
+                metadata_json,
+                b"\r\n",
+                f"--{boundary}--\r\n".encode(),
+            ]
+        )
```
jamesonuk: Could probably be tidied up.
I think
```python
data = (
    f"--{boundary}\r\n"
    'Content-Disposition: form-data; name="metadata"\r\n'
    f"Content-Length: {len(metadata_json)}\r\n"
    "\r\n"
    f"{metadata_json.decode()}\r\n"
    f"--{boundary}--\r\n"
).encode()
```
should work (coded blind though)

src/aioamazondevices/implementation/http2.py
```diff
@@ -377,6 +447,9 @@ async def _stream_and_process(self) -> None:
 
                     for part in avs_stream_parser.feed(chunk):
                         await self._handle_part(part)
+            except WrongAVSSite:
+                # Restarting with the correct AVS site
+                return
```
jamesonuk: I would maybe move this up a level to `_get_avs_directives` and add something like 
```python
        except WrongAVSSite:
            _LOGGER.debug("AVS site redirect, reconnecting to %s", self._avs_directive_site)
            continue  # reconnect immediately with no sleep
```
which would just save the delay on reconnect and is possibly a better place to handle the restart ?

In fact there is an argument that the exception handling and retry loop in `_get_avs_directives` in now possibly redundant.    We could actually push everything back to `_run_tasks` which applies an exponential backoff and restarts both ping and avs tasks (in this case the ping task won't have been setup yet anyway)

jamesonuk (dismissed):

src/aioamazondevices/implementation/http2.py
chemelli74: Tested and not working... this is why I went for the find; at least for the first testing run

src/aioamazondevices/implementation/http2.py
chemelli74: Absolutely !

src/aioamazondevices/implementation/http2.py
chemelli74: Will test

chemelli74: In the end I used httpx to build the body ;-)

src/aioamazondevices/implementation/http2.py
chemelli74: Agree, cleaned-up

src/aioamazondevices/implementation/http2.py
chemelli74:
```
  File "/workspaces/aioamazondevices/src/aioamazondevices/utils.py", line 49, in _validate_content_type
    raise ValueError(f"Unexpected content-type: {content_type!r}")
ValueError: Unexpected content-type: 'text/plain'
```

jamesonuk: Ah yes it needs the parser too.

This will properly parse it reusing the existing code.   Still a question of what we want to do in the unlikely event we get JSON back but can't get site from it
```python
        boundary = http2_parse_boundary_delimiter(
            response.headers.get("content-type", "")
        )
        if boundary is None:
            _LOGGER.warning("Missing multipart boundary in SynchronizeState response")
            return

        parser = AvsDirectiveStreamParser(boundary)
        parts = parser.feed(response.content)
        chunk_json = next(
            (http2_extract_json_from_part(p) for p in parts if p),
            None,
        )
        if chunk_json is None:
            _LOGGER.warning("Could not extract JSON from SynchronizeState response")
            return

        site = chunk_json.get("directive", {}).get("payload", {}).get("endpoint")
        # check we actually have site? or use ["directive"]["payload"]["endpoint"] and let TypeError/KeyError propagate
        self._avs_directive_site = site
```

chemelli74: A bit overkill probably, but it will run for so few users that we can go with it

PR #884, opened 2026-06-12
fix: merge sensors data updates

chemelli74:
## Summary by CodeRabbit

* **Refactor**
  * Improved sensor state handling by reusing an instance-level cache to retain per-device sensor data across calls. Cached results are refreshed for devices present in the current endpoints list, and entries for devices no longer available are removed to prevent stale sensor information.


chemelli74:
Superseded by #1080

PR #886, opened 2026-06-13
refactor: change none returns to exceptions

jamesonuk:
## Summary by CodeRabbit

* **Bug Fixes**
  * Missing or invalid multipart response boundaries are now reported as explicit errors instead of being ignored.
  * Buffer overflow conditions in directive streams now raise errors that propagate to trigger proper reconnect/recovery.
  * Token refresh failures no longer get silently swallowed, ensuring upstream retry/reconnect logic can run reliably.

### DIFF
--- a/src/aioamazondevices/implementation/http2.py
+++ b/src/aioamazondevices/implementation/http2.py
@@ -339,7 +339,7 @@ class AmazonHTTP2Client:
             REFRESH_ACCESS_TOKEN
         )
         if not refresh_successful:
-            raise CannotRetrieveData("Failed to refresh access token")
+            _LOGGER.warning("Failed to refresh access token")
 
     async def _stream_and_process(self) -> None:
         """Open stream and process incoming directives."""
--- a/src/aioamazondevices/implementation/http2.py
+++ b/src/aioamazondevices/implementation/http2.py
@@ -302,9 +302,6 @@ class AmazonHTTP2Client:
         boundary = http2_parse_boundary_delimiter(
             response.headers.get("content-type", "")
         )
-        if boundary is None:
-            _LOGGER.warning("Missing multipart boundary in SynchronizeState response")
-            return
 
         parser = AvsDirectiveStreamParser(boundary)
         parts = parser.feed(response.content)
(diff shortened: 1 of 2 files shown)


src/aioamazondevices/implementation/http2.py
```diff
@@ -330,29 +326,20 @@ async def _get_avs_directives(self) -> None:
...
-                return False
-        except (CannotConnect, CannotRetrieveData) as exc:
-            _LOGGER.warning("Failed to refresh access token: %s", exc)
-            return False
-        return True
+        refresh_successful, _ = await self._http_wrapper.refresh_data(
+            REFRESH_ACCESS_TOKEN
+        )
+        if not refresh_successful:
+            raise CannotRetrieveData("Failed to refresh access token")
```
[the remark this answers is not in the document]
jamesonuk: @coderabbitai do you consider this now resolved?    if the token update fails this will get logged but processing will continue and the next call will attempt to use the existing token.    If it has expired or there are other issues this will be raised in subsequent calls that will trigger a restart / reuath?

src/aioamazondevices/implementation/http2.py
```diff
@@ -330,29 +326,20 @@ async def _get_avs_directives(self) -> None:
         while not self._stop_event.is_set():
             self._connected_event.clear()
 
-            if not (await self._refresh_token()):
-                await asyncio.sleep(HTTP2_RECONNECT_DELAY)
-                continue
-
+            await self._refresh_token()
```
chemelli74, line 329: Not sure we should refresh token at each iteration.
Does it come with an expire date ? I think so

jamesonuk: It is actually the opposite.   We refresh the token each time we connect.    I do have a note to come back to this as the code is what is already in main and has not been touched since the original branch code.

I am not sure sure what happens with token expiry as in theory this will connect to the http2 stream after refreshing the token and that token will never refresh again (which is possibly fine for stream as once connection is established it will stay established but it could cause issues for pings)

Either way this PR doesn't change the logic of how it is handled at all and it is the same as main

chemelli74 (approved):

src/aioamazondevices/implementation/http2.py
chemelli74: I know it's the same as curremt code, just wanted to fix it while reworking ;-)

PR #885, opened 2026-06-12
refactor: use yarl URL builder

chemelli74:
## Summary by CodeRabbit

* **Refactor**
  * Standardized how request URLs are built across device operations for more consistent routing.
  * Normalized shared endpoint paths and query handling (media, notifications, history, DND, and GraphQL).
  * Improved session/token handling by using consistent URL formatting when refreshing and recording requests.
* **New Features**
  * Added support for additional authentication and welcome routes.
  * Expanded behaviors support with preview and automations endpoints.
* **Bug Fixes**
  * Improved sign-in redirect detection and corrected music providers endpoint query behavior.

PR #888, opened 2026-06-15
fix: ensure todo domain updates when changed

jamesonuk:
With current code `_base_url` is set as an immutable string when the handler is created.
With a new login the domain will likely change however this URL will always be pointing at .com resulting in a 401 error (until next time the API is created as that will pick up the updated domain.

## Summary by CodeRabbit

* **Refactor**
  * Optimized internal URL construction logic in the To-Do handler for improved code efficiency.


chemelli74:
Should be already fixed in #855

jamesonuk:
superseded by https://github.com/chemelli74/aioamazondevices/pull/885

PR #896, opened 2026-06-16
fix: avoid errors when parsing todo item changes

### LINKED ISSUE #893: fix: guard against KeyError in _handle_item_change when item_id is missing from list
## Summary

In `src/aioamazondevices/api.py`, the `_handle_item_change` method performs a direct dictionary index lookup `list_items[item_id]` after fetching the ToDo list items. If `item_id` is not present in the returned dict (e.g., due to a race condition or eventual consistency after a change event), this raises a `KeyError` and interrupts push event processing for what is otherwise a recoverable condition.

## Suggested Fix

Replace the direct index with a safe `.get()` call and handle the missing-item case gracefully:

```python
items = list_items.get(item_id)
if items is None:
    _LOGGER.debug(
        "Item %s not found in list %s after ItemChange event %s",
        item_id,
        list_id,
        list_event_type,
    )
    return
```

## References

- Identified during review of PR #891 (comment: https://github.com/chemelli74/aioamazondevices/pull/891#discussion_r3420151725)
- Requested by @chemelli74 to be tracked as a follow-up issue.

### LINKED ISSUE #894: bug: unsupported list event type guard is unreachable in _handle_item_change
## Summary

In `_handle_item_change` (introduced in PR #891), the guard that handles unsupported `eventName` values is unreachable.

## Root Cause

`AmazonListEventType(payload["eventName"])` raises a `ValueError` immediately when `payload["eventName"]` contains an unknown value. This happens **before** the `if list_event_type not in AmazonListEventType` check at the next lines can run, meaning unsupported events will bubble up an unhandled exception instead of being safely ignored.

## Relevant Code (`src/aioamazondevices/api.py`)

```python
list_event_type = AmazonListEventType(payload["eventName"])  # raises ValueError for unknown values

if list_event_type not in AmazonListEventType:  # never reached
    _LOGGER.warning("Received unsupported list event type: %s", list_event_type)
    return
```

## Expected Fix

Wrap the enum construction in a `try/except ValueError` block:

```python
raw_event_name = payload["eventName"]
try:
    list_event_type = AmazonListEventType(raw_event_name)
except ValueError:
    _LOGGER.warning("Received unsupported list event type: %s", raw_event_name)
    return

# remove the now-redundant guard:
# if list_event_type not in AmazonListEventType: ...
```

## References

- PR: #891 (comment: https://github.com/chemelli74/aioamazondevices/pull/891#discussion_r3420151717)

/cc @chemelli74

jamesonuk:
fixes: https://github.com/chemelli74/aioamazondevices/issues/894
fixes https://github.com/chemelli74/aioamazondevices/issues/893


chemelli74 (approved):

src/aioamazondevices/api.py
```diff
@@ -314,14 +314,20 @@ async def _handle_audio_player_state_event(self) -> None:
...
+
+        _LOGGER.debug("Received ItemChange for %s: %s", list_id, event_name)
 
-        if list_event_type not in AmazonListEventType:
-            _LOGGER.warning("Received unsupported list event type: %s", list_event_type)
+        try:
+            list_event_type = AmazonListEventType(event_name)
+        except ValueError:
+            _LOGGER.warning("Received unsupported list event type: %s", event_name)
             return
```
chemelli74, line 329: not sure what we are trying to fix as we alredy check if the value is known (means in the list)

src/aioamazondevices/api.py
```diff
@@ -330,7 +336,6 @@ async def _handle_item_change_event(self, payload: dict[str, Any]) -> None:
             items = list_items[item_id]
 
         list_event = AmazonListEvent(list_id, item_id, list_event_type, items=items)
-
```
chemelli74, line 333: can be reverted

src/aioamazondevices/api.py
jamesonuk: At this point `list_event_type` is not set so this is the code to set it.   There are two options.   This try/catch block which does the assignment and validates it in one go (an invalid `event_name` would raise `ValueError`).
```python
        try:
            list_event_type = AmazonListEventType(event_name)
        except ValueError:
            _LOGGER.warning("Received unsupported list event type: %s", event_name)
            return
```
Or 
```python
if event_name not in AmazonListEventType:
    _LOGGER.warning("Received unsupported list event type: %s", event_name)
    return

list_event_type = AmazonListEventType(event_name)
```

The only difference here is that there are two lookups against the enum.    Difference in performance is absolutely non-existent and I don't have a strong preference.

I will switch to the second version

chemelli74 (approved):

PR #900, opened 2026-06-17
fix: communication preferences

jamesonuk:
remove redundant values from request (avoids typing issues)
only set communication preferences if device says communications are allowed

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved communication preferences handling by refining how preferences are requested and parsed, including correctly reflecting when communications are disabled.
  * Updated device communication settings syncing so each device receives the latest communication settings (or an empty set when unavailable), including speaker-group devices.

### DIFF
--- a/src/aioamazondevices/implementation/communication.py
+++ b/src/aioamazondevices/implementation/communication.py
@@ -83,16 +83,20 @@ class AlexaCommunicationsHandler:
             resp_json = await self._http_wrapper.response_to_json(resp)
 
             device_communication_preferences: dict[str, str] = {}
-            for device_prefs in resp_json.get("devicePermissionsPreferences", {}):
+            for device_permissions_pref in resp_json.get(
+                "devicePermissionsPreferences", {}
+            ):
                 if (
-                    device_prefs.get("devicePreference") == "communications"
-                    and device_prefs.get("allowed") is False
-                ):
+                    device_prefs := device_permissions_pref.get("devicePreference")
+                ) == "communications" and device_permissions_pref.get(
+                    "allowed"
+                ) is False:
+                    # Force empty data if communications are not allowed for the device
                     device_communication_preferences = {}
                     break
-                device_communication_preferences[
-                    device_prefs.get("devicePreference")
-                ] = device_prefs.get("state")
+                device_communication_preferences[device_prefs] = (
+                    device_permissions_pref.get("state")
+                )
 
             communication_preferences[device.serial_number] = (
                 device_communication_preferences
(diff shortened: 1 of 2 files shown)


src/aioamazondevices/implementation/communication.py
```diff
@@ -84,8 +82,14 @@ async def get_communication_preferences(
             )
             resp_json = await self._http_wrapper.response_to_json(resp)
 
-            comms_preferences = {}
+            comms_preferences: dict[str, str] = {}
```
chemelli74: To be consistent with the global var:

```suggestion
            device_communication_preferences: dict[str, str] = {}
```

chemelli74: Should be part of the if so it starts blank for each device

src/aioamazondevices/implementation/communication.py
```diff
@@ -84,8 +82,14 @@ async def get_communication_preferences(
...
             for device_prefs in resp_json.get("devicePermissionsPreferences", {}):
+                if (
+                    device_prefs.get("devicePreference") == "communications"
+                    and device_prefs.get("allowed") is False
+                ):
+                    comms_preferences = {}
```
chemelli74:
```suggestion
```

If above comment is applied, this will redundant

chemelli74 (changes requested):

src/aioamazondevices/implementation/communication.py
jamesonuk: there are two levels of nesting.   The one at the top of the function is by device.
the second is over the actual preferences.

```json
{
  "devicePermissionsPreferences": [
    {
      "devicePreference": "communications",
      "allowed": true,
      "allowedStates": [],
      "state": "ON"
    },
    {
      "devicePreference": "dropin",
      "allowed": true,
      "allowedStates": [
        "All",
        "Home",
        "Off"
      ],
      "state": "Home"
    },
    {
      "devicePreference": "announcements",
      "allowed": true,
      "allowedStates": [],
      "state": "ON"
    }
  ],
  "deviceId": "xxx",
  "deviceType": "yyy"
}
```` 
but the order is not consistent so you need to iterate over the array to find the communications entry.

So setting it here means it is blank per device

src/aioamazondevices/implementation/communication.py
jamesonuk: as above, this is needed because we cannot guarantee the ordering.
If the first thing on the list is announcements and second is communications then we will have already set the announcements entry in the dict for this device.   This ensure it is empty.

chemelli74 (approved):

PR #902, opened 2026-06-18
refactor: cleanup url's leftovers

chemelli74:
## Summary by CodeRabbit

* **Refactor**
  * Improved domain and HTTPS URL handling across authentication, token refresh, and session cookie updates for more consistent requests.
  * Updated stored login “site” and Alexa host resolution to align with the retail site URL.
  * Adjusted shopping-list API endpoint construction to use the renamed base path constant for correct targeting.

### DIFF
--- a/src/aioamazondevices/http_wrapper.py
+++ b/src/aioamazondevices/http_wrapper.py
@@ -271,7 +271,7 @@ class AmazonHttpWrapper:
             "di.os.version": AMAZON_CLIENT_OS,
             "current_version": "6.12.4",
             "previous_version": "6.12.4",
-            "domain": f"www.amazon.{self._session_state_data.domain}",
+            "domain": self._session_state_data.retail_site_url.host,
         }
 
         _, raw_resp = await self.session_request(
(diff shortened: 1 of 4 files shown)


src/aioamazondevices/http_wrapper.py
```diff
@@ -342,7 +342,10 @@ async def session_request(
             else self._cookies
         )
         self._session.cookie_jar.update_cookies(
-            _cookies, URL(f"amazon.{self._session_state_data.domain}")
+            _cookies,
+            URL.build(
+                scheme="https", host=f".amazon.{self._session_state_data.domain}"
+            ),
         )
```
jamesonuk, line 349: first of there is actually a `set_cookies` function which does this and is used when logging in (see below as you have changed the call to it from `_refresh_auth_cookies`).

I do have a small concern as this changes how things are handled.   Not having a scheme before was a bug but it meant aiothttp basically ignored the domain and stored it as a session cookie and passed it to everything.
I have done some testing and it seems to be working but something to bear in mind.

jamesonuk (changes requested):
Probably change to use the `set_cookies` function seems as it exists but otherwise looks good and limited testing shows it seems to be working fine.

Just need to be aware that this will fix a bug in existing code but that will change how cookies are handled.   Don't think it will make a difference but worth noting.

src/aioamazondevices/http_wrapper.py
chemelli74: Good spot, updated.

jamesonuk (approved):

PR #906, opened 2026-06-19
refactor: move raw media command

chemelli74:
## Release Notes

* **Refactor**
  * Restructured internal media control request handling to improve code organization and maintainability. Stop commands continue routing through the existing Alexa command sequence, while other media control commands now route through an optimized handler. Complete backward compatibility maintained; no changes to user-facing behavior.


jamesonuk (approved):
I cannot test as don't have a dev machine available but the code looks good.   I presume this get merged at the same time we split stuff out as not sure how this ended in here anyway....

the library test should test stop, pause and play commands so if that works it is good to go

PR #911, opened 2026-06-22
fix: only return 'allowed' comms preferences

jamesonuk:
Caters for scenarios such as fire TV stick which supports comms and announcements but not dropin.

## Bug Fixes
* Fixed communication preference parsing and handling for Amazon devices. The system now correctly extracts and filters individual permission entries based on their specific allowance status, rather than clearing all preferences when communication restrictions are detected. This ensures more accurate preservation, processing, and display of user communication preference settings while respecting permission configurations.

### DIFF
--- a/src/aioamazondevices/implementation/communication.py
+++ b/src/aioamazondevices/implementation/communication.py
@@ -86,17 +86,12 @@ class AlexaCommunicationsHandler:
             for device_permissions_pref in resp_json.get(
                 "devicePermissionsPreferences", {}
             ):
-                if (
-                    device_prefs := device_permissions_pref.get("devicePreference")
-                ) == "communications" and device_permissions_pref.get(
-                    "allowed"
-                ) is False:
-                    # Force empty data if communications are not allowed for the device
-                    device_communication_preferences = {}
-                    break
-                device_communication_preferences[device_prefs] = (
-                    device_permissions_pref.get("state")
-                )
+                device_pref = device_permissions_pref.get("devicePreference")
+                pref_state = device_permissions_pref.get("state")
+                pref_allowed = device_permissions_pref.get("allowed")
+
+                if pref_allowed is True:
+                    device_communication_preferences[device_pref] = pref_state
 
             communication_preferences[device.serial_number] = (
                 device_communication_preferences
--- a/src/aioamazondevices/implementation/communication.py
+++ b/src/aioamazondevices/implementation/communication.py
@@ -86,11 +86,11 @@ class AlexaCommunicationsHandler:
             for device_permissions_pref in resp_json.get(
                 "devicePermissionsPreferences", {}
             ):
-                device_pref = device_permissions_pref.get("devicePreference")
+                device_pref = device_permissions_pref["devicePreference"]
                 pref_state = device_permissions_pref.get("state")
                 pref_allowed = device_permissions_pref.get("allowed")
 
-                if device_pref and pref_allowed is True:
+                if pref_allowed is True:
                     device_communication_preferences[device_pref] = pref_state
 
             communication_preferences[device.serial_number] = (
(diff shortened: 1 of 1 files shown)


src/aioamazondevices/implementation/communication.py
```diff
@@ -86,17 +86,12 @@ async def get_communication_preferences(
...
-                    break
-                device_communication_preferences[device_prefs] = (
-                    device_permissions_pref.get("state")
-                )
+                device_pref = device_permissions_pref.get("devicePreference")
+                pref_state = device_permissions_pref.get("state")
+                pref_allowed = device_permissions_pref.get("allowed")
+
+                if pref_allowed is True:
+                    device_communication_preferences[device_pref] = pref_state
```
chemelli74, line 96: device_pref can be None

jamesonuk: it is the key field so massively unlikely but will skip if None

chemelli74: Agree, but then don't use get while assigning the value ;-)

src/aioamazondevices/implementation/communication.py
```diff
@@ -86,17 +86,12 @@ async def get_communication_preferences(
...
-                    break
-                device_communication_preferences[device_prefs] = (
-                    device_permissions_pref.get("state")
-                )
+                device_pref = device_permissions_pref["devicePreference"]
+                pref_state = device_permissions_pref.get("state")
+                pref_allowed = device_permissions_pref.get("allowed")
+
+                if pref_allowed is True:
+                    device_communication_preferences[device_pref] = pref_state
```
chemelli74, line 96:
```suggestion
                if (
                    not device_prefs := device_permissions_pref.get("devicePreference")
                ) or device_permissions_pref.get(
                    "allowed"
                ) is False:
                    # Force empty data if no data are retrieved or 
                    # communications are not allowed for the device
                    device_communication_preferences = {}
                    break
                device_communication_preferences[device_prefs] = (
                    device_permissions_pref.get("state")
                )
```

chemelli74 (changes requested):

src/aioamazondevices/implementation/communication.py
jamesonuk: You cannot do this.   `device_communication_preferences` is per device.

This is a the JSON from a speaker group
```json
{
  "devicePermissionsPreferences": [
    {
      "devicePreference": "dropin",
      "allowed": false,
      "allowedStates": [
        "All",
        "Home",
        "Off"
      ],
      "state": null
    },
    {
      "devicePreference": "announcements",
      "allowed": false,
      "allowedStates": [],
      "state": null
    },
    {
      "devicePreference": "communications",
      "allowed": false,
      "allowedStates": [],
      "state": null
    }
  ],
  "deviceId": "xxx",
  "deviceType": "A3C9PE6TNYLTCH"
}
```
so `device_communication_preferences` should be {} (nothing is allowed)
but this is a Fire TV Stick
```json
{
  "devicePermissionsPreferences": [
    {
      "devicePreference": "communications",
      "allowed": true,
      "allowedStates": [],
      "state": "ON"
    },
    {
      "devicePreference": "dropin",
      "allowed": false,
      "allowedStates": [
        "All",
        "Home",
        "Off"
      ],
      "state": null
    },
    {
      "devicePreference": "announcements",
      "allowed": true,
      "allowedStates": [],
      "state": "ON"
    }
  ],
  "deviceId": "xxx",
  "deviceType": "ADVBD696BHNV5"
}
```
so communications and announcements are allowed but dropin is not.

This should have
```
"device_communication_preferences:" {
  "announcements": "ON",
  "communications": "OFF"
}
```
and it should have announcements and communications controls in HA.

With the proposed change `device_communication_preferences`  would be {} and no controls would be created in HA

chemelli74 (approved):

PR #921, opened 2026-06-24
refactor: logging

chemelli74:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when loading account, device communication, history, routines, and to-do data.
  * Fixed several response-parsing paths so app features now read the correct data sections more consistently.
  * Strengthened request handling during setup and login flows for more dependable behavior.


src/aioamazondevices/implementation/todo.py
```diff
@@ -57,9 +57,7 @@ async def _call_lists_api(
     async def update_lists(self) -> None:
         """Fetch all available Alexa shopping lists and stores it."""
         raw_resp = await self._call_lists_api(HTTPMethod.POST, "fetch")
-        response_json = await self._http_wrapper.response_to_json(
-            raw_resp, "listInfoList"
-        )
+        response_json = await self._http_wrapper.response_to_json(raw_resp, "fetch")
```
jamesonuk: Perhaps using something more descriptive than fetch?  `todo (fetch)` or something?
Will be easy to trace based on the contents but if the point is to make logging clearer we should possibly use something more descriptive

src/aioamazondevices/implementation/todo.py
```diff
@@ -81,7 +79,9 @@ async def get_list_items(
             query={"limit": limit},
         )
 
-        response_json = await self._http_wrapper.response_to_json(raw_resp)
+        response_json = await self._http_wrapper.response_to_json(
+            raw_resp, "fetch(items)"
```
jamesonuk: Same as above.    Perhaps something more descriptive?

jamesonuk (approved):
Looks good.
Just a tiny question on whether we should use some more descriptive names but it should be obvious what it is from the previous logged line and payload.

src/aioamazondevices/implementation/todo.py
chemelli74: Yeah I was thinking the same as well.
Will update

src/aioamazondevices/implementation/todo.py
chemelli74: What about `todo(items/fetch)` ?

chemelli74:
> Looks good.
> Just a tiny question on whether we should use some more descriptive names but it should be obvious what it is from the previous logged line and payload.

Open to suggestions

jamesonuk:
> > Looks good.
> > Just a tiny question on whether we should use some more descriptive names but it should be obvious what it is from the previous logged line and payload.
> 
> Open to suggestions

`todo (fetch)`
`todo (fetch/items)` or todo (items)`

work.   I guess as long as it has some mention of todo and what it is doing the name is less important as it is only used in logging but will be a lot more obvious than just fetch.

jamesonuk (approved):

PR #925, opened 2026-06-25
fix: bypass comms unavailable error

jamesonuk:
I don't particularly like it but currently users are unable to set up or start the integration at all.
I think as a temporary measure we should bypass the error and see if this does go away or whether we need to do something different

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when refreshing communication settings for devices.
  * If a device’s settings can’t be retrieved, the app now skips that device and continues processing the rest instead of failing the whole refresh.
  * Updated communication preferences are now kept more consistently across refreshes.

### DIFF
--- a/src/aioamazondevices/implementation/communication.py
+++ b/src/aioamazondevices/implementation/communication.py
@@ -85,7 +85,8 @@ class AlexaCommunicationsHandler:
                 )
             except CannotRetrieveData:
                 _LOGGER.warning(
-                    "Failed to refresh communications settings for device %s, used cached values.", device.account_name
+                    "Failed to refresh communications settings for device %s, used cached values.",  # noqa: E501
+                    device.account_name,
                 )
                 continue
             resp_json = await self._http_wrapper.response_to_json(
--- a/src/aioamazondevices/implementation/communication.py
+++ b/src/aioamazondevices/implementation/communication.py
@@ -62,7 +62,7 @@ class AlexaCommunicationsHandler:
         self, devices: list[AmazonDevice]
     ) -> dict[str, dict[str, str]]:
         """Get communication preferences for a device."""
-        communication_preferences = {}
+        communication_preferences: dict[str, dict[str, str]] = {}
         for device in devices:
             query_string = {
                 "devicePreferences": [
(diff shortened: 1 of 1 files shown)


src/aioamazondevices/implementation/communication.py
```diff
@@ -77,9 +79,17 @@ async def get_communication_preferences(
...
+            try:
+                _, resp = await self._http_wrapper.session_request(
+                    method=HTTPMethod.GET, url=url
+                )
+            except CannotRetrieveData as e:
+                if str(e) == "Request failed: Service Unavailable":
+                    _LOGGER.debug("unable to get comms preferences")
+                    # no point continuing to try other devices as they
+                    # will also fail.   Just return what we have so far
+                    return communication_preferences
```
chemelli74: I found that often I get a failure only for 1 device, not for all:

```suggestion
                if "Service Unavailable" in str(exc):
                    # If we fail to get the preferences
                    # we will return an empty dict for the device
                    _LOGGER.warning(
                        "Failed to get communication preferences for device %s",
                        device.serial_number,
                    )
                    communication_preferences[device.serial_number] = {}
                    continue
```

chemelli74 (changes requested):

src/aioamazondevices/implementation/communication.py
chemelli74: Also we should evaluate what to return as `{}` will make entities to be added/removed at every status change.

I would rather assume that they are all `ON` when you cannot retrieve them to avoid this hula-hop

src/aioamazondevices/implementation/communication.py
```diff
@@ -77,9 +79,15 @@ async def get_communication_preferences(
...
-            _, resp = await self._http_wrapper.session_request(
-                method=HTTPMethod.GET, url=url
-            )
+            try:
+                _, resp = await self._http_wrapper.session_request(
+                    method=HTTPMethod.GET, url=url
+                )
+            except CannotRetrieveData:
+                _LOGGER.warning(
+                    "Failed to refresh communications settings, maybe stale."
```
chemelli74:
```suggestion
                    "Failed to refresh communications settings for device %s, used cached values.", device.account_name
```

chemelli74 (approved):

PR #928, opened 2026-06-26
fix: skip comms calls for devices that cannot be controlled

jamesonuk:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved device handling so communication preferences are no longer fetched for unsupported speaker group and AQM device types.
  * Prevented these devices from being added to communication preference results, reducing unnecessary processing and incorrect updates.

### DIFF
--- a/src/aioamazondevices/implementation/communication.py
+++ b/src/aioamazondevices/implementation/communication.py
@@ -4,6 +4,7 @@ from http import HTTPMethod
 
 from yarl import URL
 
+from aioamazondevices.const.devices import DEVICE_TYPE_AQM, SPEAKER_GROUP_FAMILY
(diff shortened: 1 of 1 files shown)


src/aioamazondevices/implementation/communication.py
```diff
@@ -64,6 +65,12 @@ async def get_communication_preferences(
     ) -> dict[str, dict[str, str]]:
         """Get communication preferences for a device."""
         for device in devices:
+            if (
+                device.device_family == SPEAKER_GROUP_FAMILY
+                or device.device_type == DEVICE_TYPE_AQM
+            ):
+                continue
```
chemelli74, line 73:
```suggestion
                # Those devices cannot be controlled, avoiding one call
                continue
```

chemelli74 (approved):

PR #933, opened 2026-06-27
fix: log connection error once per comms settings loop

jamesonuk:


src/aioamazondevices/implementation/communication.py
```diff
@@ -116,4 +114,9 @@ async def get_communication_preferences(
                 device_communication_preferences
             )
 
+        if failed_updates:
+            _LOGGER.warning(
+                "Failed to refresh communications settings for [%s], will use cached values.",  # noqa: E501
```
chemelli74, line 119: I would love to know if cached data is empty or not for the failing device.

PR #975, opened 2026-07-17
fix: update cluster volume on member volume update

jamesonuk:
Have not had chance to test this properly but looking at the app the volume of a speaker group is the average volume of all the cluster members (eg. if you have a two speaker group and set one to 30 and the other to 40 then the group will show as volume 35, set one to 30 and the other to 50 then the group shows as 40.

This should address https://github.com/home-assistant/core/issues/174422 but just needs testing.   Also need to check the additional field on `AmazonDevice` doesn't break the test snapshots on HA side

## Summary by CodeRabbit

* **Bug Fixes**
  * Speaker-group volume levels now stay synchronized when a member device’s volume changes.
  * Group volume is calculated from available member volumes, with mute status reflected accurately.
  * Devices associated with speaker groups are now correctly recognized for volume updates.
  * Groups without available member volume data are no longer incorrectly updated.
  * Volume handling remains consistent for devices discovered through different device sources.


src/aioamazondevices/api.py
```diff
@@ -353,7 +305,21 @@ async def _handle_volume_change_event(self, payload: dict[str, Any]) -> None:
             volume = AmazonVolumeState(
                 payload.get("volumeSetting"), payload.get("isMuted")
             )
+            # check if device is part of a cluster
             self._media_handler.update_cached_device_volume(serial, volume)
+            volume_device: AmazonDevice | None = self._device_handler.devices.get(
+                serial, None
```
```suggestion
                serial
```

src/aioamazondevices/structures.py
```diff
@@ -64,6 +49,7 @@ class AmazonDevice:
     device_owner_customer_id: str
     household_device: bool
     device_cluster_members: dict[str, str | None]
+    parent_clusters: list[str] | None
```
chemelli74:
```suggestion
    parent_clusters: list[str]
```

src/aioamazondevices/implementation/device.py
```diff
@@ -97,6 +94,7 @@ async def get_base_devices(self) -> None:
                 device_cluster_members=dict.fromkeys(
                     device["clusterMembers"] or [serial_number]
                 ),
+                parent_clusters=device.get("parentClusters"),
```
chemelli74:
```suggestion
                parent_clusters=device.get("parentClusters", []),
```

src/aioamazondevices/implementation/device.py
```diff
@@ -246,6 +244,7 @@ async def _get_devices_endpoint_data(self) -> dict[str, dict[str, Any]]:
                 or "n/a",
                 household_device=False,
                 device_cluster_members={aqm_serial_number: DEVICE_TYPE_AQM},
+                parent_clusters=None,
```
chemelli74:
```suggestion
                parent_clusters=[],
```

src/aioamazondevices/api.py
```diff
@@ -353,7 +305,21 @@ async def _handle_volume_change_event(self, payload: dict[str, Any]) -> None:
...
+            )
+            if volume_device and volume_device.parent_clusters:
```
```suggestion
            if volume_device:
```

src/aioamazondevices/api.py
```diff
@@ -353,7 +305,21 @@ async def _handle_volume_change_event(self, payload: dict[str, Any]) -> None:
...
+                for parent_serial in volume_device.parent_clusters:
+                    parent_cluster: AmazonDevice | None = (
+                        self._device_handler.devices.get(parent_serial, None)
+                    )
+                    if parent_cluster:
```
```suggestion
                    if (parent_cluster :=  (
                        self._device_handler.devices.get(parent_serial, None)
                    ):
```

chemelli74 (changes requested):
A few optimizations

src/aioamazondevices/implementation/device.py
```diff
@@ -97,6 +97,7 @@ async def get_base_devices(self) -> None:
                 device_cluster_members=dict.fromkeys(
                     device["clusterMembers"] or [serial_number]
                 ),
+                parent_clusters=device.get("parentClusters", []) or [],
```
chemelli74:
```suggestion
                parent_clusters=device.get("parentClusters") or [],
```

chemelli74 (changes requested):
One very tiny change and I think we are good to go.

chemelli74 (approved):
LGTM !

chemelli74:
@coderabbitai, full review please

PR #992, opened 2026-07-24
fix: ensure sync history sends update

jamesonuk:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved consistency when vocal history is refreshed after device state changes.
  * Ensured manual and automatic history synchronization use the same refresh behavior.

* **Behavior Changes**
  * History synchronization now completes without returning history data directly.


jamesonuk:
Not necessary

PR #1021, opened 2026-08-08
fix: Recover vocal history for devices that report no deviceInfo

MasterDevwi:
## Summary

Amazon returns vocal history records with a null `deviceInfo` for some devices, notably Sonos speakers with Alexa built in. Those records are discarded, so no vocal record ever reaches the consumer and voice history never works on those devices.

On my account every captured record had `deviceInfo: None`, including ordinary `GENERAL` voice commands, so nothing was ever surfaced.

### Recover the serial number from `activityKey`

When `deviceInfo` is missing, the serial number is still carried by `activityKey`, which Amazon formats as `<customerId>#<timestamp>#<deviceType>#<deviceSerialNumber>`.

The parser now falls back to that value, and accepts it only when it belongs to a known device. Because it can only resolve to an already-known serial, it cannot reintroduce the `KeyError` that the current `deviceInfo is None` guard protects against.

I verified this against real captured data: all eight distinct serials parsed from `activityKey` matched devices the integration had already discovered.

### Refresh history from volume changes on Sonos

Sonos speakers never emit `PUSH_EQUALIZER_STATE_CHANGE`, which is the only push that triggers a history refresh. Across several debug sessions I recorded zero equalizer pushes and zero `PUSH_ACTIVITY`.

They do reliably emit `PUSH_VOLUME_CHANGE` when Alexa ducks audio to answer. That is now routed into the same history refresh, but only for devices whose `manufacturer` identifies them as Sonos, so Echo devices are unaffected and keep using the equalizer event.

A short debounce is applied because wake word arbitration makes several nearby speakers report a volume change for a single utterance.

### Ignore false wake word records

`FALSE_WAKE_WORD_1P` records carry no transcript, so they surface as a vocal record with an empty title. Since only the latest record per device is kept, an accidental wake also hides that device's real last command until it is used again.

This is a separate commit and can be dropped independently if you would rather not change behaviour for Echo devices.

## Verification

Unit tests cover the activity key fallback, the known device check, malformed keys, the Sonos gate, the debounce, and the false wake filter. The full suite passes, along with ruff and the formatter.

I also ran this against a live installation with roughly a dozen Sonos speakers. Speaking to one produced:

```
12:59:20.344  Detected push event type <PUSH_VOLUME_CHANGE> on device <redacted>
12:59:23.358  Emitting vocal history event to subscribers
12:59:23.361  entity updated
```

The resulting record was correct despite `deviceInfo` being null:

```
voice_command = 'alexa what time is it'
voice_reply   = "It's 12:59 PM."
intent        = 'WhatTimeIntent'
```

The false wake change also had a visible effect. One speaker previously reported an empty command, and once its stale false wake was filtered its genuine last command surfaced instead:

```
voice_command = 'alexa set a timer for five minutes'
voice_reply   = '2nd timer, 5 minutes, starting now'
intent        = 'SetNotificationIntent'
```

False wake records were still present in the raw payload during that run, so they are being filtered rather than simply absent.

## Notes

@chemelli74 a few things worth your judgement:

1. The Sonos check uses `manufacturer`, which reports `Sonos, Inc.` on every one of my speakers, rather than a device type allow list. Happy to switch if you prefer an explicit list.
2. I renamed `_handle_eq_event_as_history_proxy` to `_refresh_vocal_history`, since two different pushes now reach it. Easy to revert if you would rather keep the original name.
3. The debounce interval is currently three seconds, chosen from observed arbitration bursts. Happy to tune it.
4. The false wake commit is separable, as noted above.

## Summary by CodeRabbit

* **New Features**
  * Improved vocal-history updates for Sonos speakers with Alexa built in.
  * Volume changes on supported Sonos devices now refresh vocal history.
  * Equalizer changes continue to refresh history.
  * Device preferences are now loaded and saved automatically.
  * Default devices can be selected and updated asynchronously.

* **Bug Fixes**
  * Prevented repeated volume events from triggering excessive history refreshes.
  * Improved handling of incomplete, malformed, and false-wake history records.
  * Improved device matching and handling when no online device is available.


MasterDevwi:
## Volume change now goes through a dedicated method

The volume case calls a dedicated method whose only job is to decide whether this is a Sonos and, if so, trigger the history update:

```python
case AmazonPushMessage.VolumeChange.value:
    await self._handle_volume_change_event(payload)
    await self._handle_volume_event_as_history_proxy(payload)
```

```python
async def _handle_volume_event_as_history_proxy(self, payload: dict[str, Any]) -> None:
    serial = payload.get("dopplerId", {}).get("deviceSerialNumber")
    if not self._is_sonos_device(serial):
        return
    ...
```

`_handle_volume_change_event` keeps doing only volume state, so nothing changes there.

## Echo behaviour is left as it is

I reverted the changes I had made to the shared paths in the first version. `_handle_eq_event_as_history_proxy` and `sync_history_state` are now identical to main and appear as context in the diff rather than as modifications. In particular I had renamed the equalizer proxy, and that rename is undone.

The only change to the existing push handler is the single added line in the volume case, so an Echo never reaches any of the new code.

## One thing I would like your call on

The Sonos path passes the set of known device serials into `get_vocal_history()`, so a serial recovered from `activityKey` is only accepted when it matches a real device. The equalizer and startup paths call it with no argument and keep their original behaviour.

That leaves the two paths slightly inconsistent. I kept it this way to avoid touching anything on the Echo side, but I am happy to either apply it everywhere or drop the parameter altogether. The parameter is optional and defaults to `None`, and an unmatched serial is harmless either way since it simply never corresponds to a device. Let me know which you prefer.

## Verification

The full test suite, ruff and the formatter all pass. The rework is a restructuring rather than a behaviour change, so the earlier live results still apply: on my installation a Sonos volume push triggers the refresh about three seconds after speaking, and the record resolves correctly despite `deviceInfo` being null. Happy to re-run a live capture if useful.

src/aioamazondevices/api.py
```diff
@@ -312,10 +319,47 @@ async def _handle_volume_change_event(self, payload: dict[str, Any]) -> None:
...
+        if device is None or not device.manufacturer:
+            return False
+
+        return device.manufacturer.lower().startswith(SONOS_MANUFACTURER_PREFIX)
+
     async def _handle_eq_event_as_history_proxy(self) -> None:
         vocal_history = await self._history_handler.get_vocal_history()
         await self._emit_history_event(vocal_history)
 
+    async def _handle_volume_event_as_history_proxy(
```
```suggestion
    async def _handle_sonos_volume_event_as_history_proxy(
```

src/aioamazondevices/api.py
```diff
@@ -51,6 +52,10 @@
 )
 from .utils import _LOGGER, scrub_fields
 
+# Wake word arbitration makes several speakers report a volume change for a
+# single utterance, so collapse bursts into one history refresh
+HISTORY_REFRESH_MIN_INTERVAL = timedelta(seconds=3)
```
```suggestion
# Wake word arbitration makes several speakers report a volume change for a
# single utterance, so collapse bursts into one history refresh
SONOS_HISTORY_REFRESH_MIN_INTERVAL = timedelta(seconds=3)
```

and move it to const/devices.py please

src/aioamazondevices/api.py
```diff
@@ -142,6 +147,7 @@ def __init__(
         initial_time = datetime.now(UTC) - timedelta(days=2)
         self._last_daily_refresh: datetime = initial_time
         self._last_endpoint_refresh: datetime = initial_time
+        self._last_history_refresh: datetime = initial_time
```
```suggestion
        self._last_sonos_history_refresh: datetime = initial_time
```

src/aioamazondevices/implementation/history.py
```diff
@@ -24,6 +24,43 @@
 
 BACKEND_REFRESH_WAIT_SECONDS = 2
 
+# Amazon formats activityKey as:
+# <customerId>#<timestamp>#<deviceType>#<deviceSerialNumber>
+ACTIVITY_KEY_PARTS = 4
+ACTIVITY_KEY_SERIAL_INDEX = 3
+
+UTTERANCE_TYPES_TO_SKIP = [
```
chemelli74: Don't we already have such a list ?

chemelli74 (changes requested):

MasterDevwi:
All four addressed and force pushed.

**Renamed the volume proxy** to `_handle_sonos_volume_event_as_history_proxy`, so the Sonos scope is obvious from the name.

**Renamed the timestamp** to `_last_sonos_history_refresh` to match.

**Renamed and moved the interval.** It is now `SONOS_HISTORY_REFRESH_MIN_INTERVAL` in `const/devices.py`, next to `SONOS_MANUFACTURER_PREFIX`.

**On the utterance list:** no, there was no existing list. It was an inline literal inside `get_vocal_history()`:

```python
if (
    utterance_type
    in [
        "ASR_TIMEOUT",
        "DEVICE_ARBITRATION",
        "NO_EXPRESSED_INTENT",
        "WAKE_WORD_ONLY",
    ]
    ...
```

I had only lifted it into a named constant so the second commit could extend it. Following the same reasoning as the interval, I have now moved it out to `const/metadata.py` as `UTTERANCE_TYPES_TO_SKIP`, and the tests import it from there rather than duplicating the values.

I put it in `metadata.py` rather than `devices.py` since it describes Amazon's utterance vocabulary rather than a device attribute, but happy to move it if you would rather keep it beside the other one.

The two commits are still independent: the second one is now a single line in `const/metadata.py` plus its regression test, so it can still be dropped on its own if you would rather not change this for Echo devices.

Full suite, ruff and the formatter all pass.

MasterDevwi:
Rebased this onto the latest `main` with no conflicts. The updated branch passes all 53 tests, Ruff checks, and formatting locally.

src/aioamazondevices/api.py
```diff
@@ -375,13 +381,50 @@ async def _handle_volume_change_event(self, payload: dict[str, Any]) -> None:
...
+        on their own and keep using the equalizer event untouched.
+        """
+        serial = payload.get("dopplerId", {}).get("deviceSerialNumber")
+        if not self._is_sonos_device(serial):
+            return
+
+        now = datetime.now(UTC)
+        if now - self._last_sonos_history_refresh < SONOS_HISTORY_REFRESH_MIN_INTERVAL:
+            _LOGGER.debug("Skipping history refresh for device %s: too soon", serial)
+            return
```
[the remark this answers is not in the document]
MasterDevwi: Thanks for flagging this. The leading-edge debounce is intentional here: one Sonos utterance produces a burst of volume events, and `get_vocal_history()` already waits for the Amazon backend to settle before fetching. Scheduling a trailing refresh would make the normal path perform two history requests for the same utterance without a distinct signal that new history is available. The burst test verifies that these events collapse into one request. I’d prefer to keep this as-is unless we observe records arriving later than the existing backend wait.

src/aioamazondevices/api.py
```diff
@@ -375,13 +381,50 @@ async def _handle_volume_change_event(self, payload: dict[str, Any]) -> None:
...
+
+        self._last_sonos_history_refresh = now
+        vocal_history = await self._history_handler.get_vocal_history(
+            set(self._device_handler.devices)
+        )
```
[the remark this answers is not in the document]
MasterDevwi: Good catch. Added the early subscriber guard in 1005242, with a regression test confirming that neither device detection nor history retrieval runs when there are no history subscribers.

MasterDevwi:
The additional `default_device` validation finding looks valid, but it is pre-existing and outside the Sonos history changes in this PR. I’m keeping that separate so this fix remains focused.

PR #1034, opened 2026-08-15
fix: improve online/offline state

### LINKED ISSUE #1063: ecobee Smart Thermostat Premium (AA1IN44SS3X6O): missing `reachability` forces `online=False`, all entities permanently unavailable
**Describe the bug**

On `aioamazondevices` 15.1.0 (Home Assistant 2026.9.1, `alexa_devices` integration), an ecobee Smart Thermostat Premium — device type `AA1IN44SS3X6O`, reported as a known device in #605 — produces 12 entities in Home Assistant and **every one of them is permanently `unavailable`**. It has been unavailable for 100% of Home Assistant's recorder retention on every occasion I have measured it, since the integration was set up in July 2026.

The cause is in `implementation/sensor.py`:

```python
if sensors:
    device.sensors = sensors
    if reachability_sensor := sensors.get("reachability"):
        device.online = reachability_sensor.value == "OK"
    else:
        device.online = False        # <-- this device lands here
else:
    device.online = False
    for device_sensor in device.sensors.values():
        device_sensor.error = True
```

This device does not report a `reachabilityStatusValue` (the `connectivity` capability mapped in `const/metadata.py` and requested in `const/queries.py`), so `online` is unconditionally `False`, and Home Assistant renders the whole device — media_player, both notify entities, connectivity, DND switch, the alarm/reminder/timer sensors — as `unavailable`.

This looks like the gap you flagged yourself when designing this in #647:

> I know a fire TV stick I have that is not connected is returning an error of `NOT_FOUND` which should really be treated as offline rather than unavailable

The implementation collapses "capability absent" and "capability present but not `OK`" into the same `online = False`. But "this device never reports reachability" is not the same claim as "this device is unreachable", and for this device type the difference is load-bearing: in #605 the reporter confirmed that `notify.<device>_speak` **works** on this exact device type ("Speak yes"). Forcing the device offline therefore removes a TTS target that is actually functional.

**To Reproduce**

1. Have an ecobee Smart Thermostat Premium (`AA1IN44SS3X6O`, Alexa built-in) on the Amazon account.
2. Set up the Home Assistant `alexa_devices` integration.
3. All 12 entities for that device are created, and all are `unavailable` — permanently, never briefly available.

**Additional context**

Either of these would resolve it:

1. Distinguish "the device exposes no `connectivity` capability at all" from "`reachability` is present but not `OK`". A device that never exposes the capability could retain the `URI_DEVICES` online value (the pre-#647 behaviour) instead of being forced offline.
2. If this device type is considered genuinely unsupported, add `AA1IN44SS3X6O` to `DEVICE_TYPES_TO_IGNORE` in `const/devices.py`, next to the ecobee4 `A18BI6KPKDOEI4` (#199). That would at least stop it from creating 12 dead entities. This seems like the worse option if speak/announce do in fact work on it.

I have not attached a raw payload because `aioamazondevices` debug logging emits live Amazon credentials into the log. I am happy to capture and redact the relevant device fragment if that would help.

Versions: `aioamazondevices` 15.1.0, Home Assistant Core 2026.9.1, Python 3.14.

chemelli74:
fixes #1063

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved devices’ last-known online status when endpoint or reachability data is unavailable.
  * Prevented missing sensor data from incorrectly marking devices as offline.
  * Reduced repeated warnings when reachability information is unavailable.
  * Improved handling of sensor state updates across devices.


src/aioamazondevices/implementation/sensor.py
```diff
@@ -55,14 +55,14 @@ async def update_sensor_data(
...
             sensors = devices_sensors.get(device.serial_number, {})
+            _LOGGER.debug(
+                "Sensors data for device %s: %s", device.serial_number, sensors
+            )
             if sensors:
                 device.sensors = sensors
                 if reachability_sensor := sensors.get("reachability"):
                     device.online = reachability_sensor.value == "OK"
-                else:
-                    device.online = False
```
jamesonuk, line 63: Might need some thought here.    online status is set from rest call to `api/devices-v2/device` which is only updated at startup and then every 24 hours.   I am presuming that value is correct.

I guess Sonos devices are not reporting a reachability sensor but if a device was offline at startup and turned on five minutes later it will not show up in HA for a whole 24 hours.

If an Amazon device didn't return the `reachability` sensor when it had before that could also lead to stale values.

Is there any sort of capability or other way of identify which devices should have a `reachability` sensor and any other way we can update the online status of Sonos devices more than once every 24 hours.

chemelli74: I don't get why if reachability is missing we force offline status.
if it's there then you get the current status, if not then you need to rely on some other field.

I found only the solution above for Sonos.
Please, as per our discord chat, chat with the affected user so you may be able to find a better solution than I did.

chemelli74: @jamesonuk, updated with some tweaks.

Please review, I would like to release the fix asap.

jamesonuk: Let me have a think about this.    I guess there are two things here.
1. devices stop returning an online (`reachability`) sensor but we expect them to.
2. devices just aren't returning an online status at all

The existing code I think was to cater for (1).   If they are not reporting as online then this should mean they aren't communicating with Amazon so it would be reasonable to assume they are offline.
(2) is more complicated as there is no way to tell whether we are expecting an online status or not.

I am not sure this does fix things.   If we do the daily poll at 07:00 and a Sonos device is offline then the rest call will flag it as offline.   If it is then turned on at 07:05 we would not pick it up as being back online until 07:00 the next day (and if it is only turned on during the day we would never see it as online).   So picking the previous status might not help in that scenario.   Similarly if Amazon devices aren't returning a `reachability` value via GraphQL then that makes me think they aren't actually online but we could keep them showing as online and available?

The simplest solution would be to revert online status to a poll of the devices endpoint, which isn't the end of the world.   That would I think give us a working value (with five minute latency) for all devices but is something we have tried to avoid.

Let me review some logs and bug reports

PR #1040, opened 2026-08-21
feat: add Amazon thermostat support

### LINKED ISSUE #576: Add support for the Amazon thermostat
**Is your feature request related to a problem? Please describe.**
I have an Amazon thermostat that only works in Alexa, unless additional hardware is used (homebridge). So, no automation can be done in Home Assistant utilizing the thermostat. Also, the current settings and schedules are unavailable in HA.

**Describe the solution you'd like**
If possible, I would like to see the thermostat show up as a device like the Echo devices do, and have the ability to at least toggle power, switch modes (cool, heat, auto, etc), manually change temperature, etc. A stretch goal would be to be able to update the schedule.

**Additional context**
None.

chemelli74:
```yaml
AmazonDeviceSensor(name='thermostat', value={
'temperatureScale': 'FAHRENHEIT',
'upperSetpoint': 72.0,
'targetSetpoint': 62.0,
'lowerSetpoint': 65.0,
'thermostatMode': 'AUTO',
'supportedModes': ['HEAT', 'COOL', 'AUTO', 'OFF'],
'heatingMinimumTemperature': 40.0,
'heatingMaximumTemperature': 87.0,
'coolingMinimumTemperature': 43.0,
'coolingMaximumTemperature': 90.0,
'temperature': 72.0}
```

fixes #576

### DIFF
--- a/src/aioamazondevices/implementation/device.py
+++ b/src/aioamazondevices/implementation/device.py
@@ -10,6 +10,7 @@ from yarl import URL
 
 from aioamazondevices.const.devices import (
     DEVICE_TYPE_AQM,
+    DEVICE_TYPE_THERMOSTAT,
     DEVICE_TYPES_HARDCODED_METADATA,
     DEVICE_TYPES_TO_IGNORE,
 )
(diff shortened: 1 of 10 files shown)


src/aioamazondevices/implementation/device.py
```diff
@@ -216,53 +217,77 @@ async def _get_devices_endpoint_data(self) -> dict[str, dict[str, Any]]:
...
+                )
             ):
-                _LOGGER.debug(
-                    "Skipping non-Amazon Air Quality Monitor: %s",
-                    aqm_endpoint,
+                continue
+            devices_endpoints[serial_number] = aqm_endpoint
+            self._endpoints[aqm_endpoint["endpointId"]] = serial_number
+
+        for therm_endpoint in data.get("thermostats", {}).get("endpoints", {}):
```
colindembovsky, line 233: Thermostat-only accounts cannot reach this discovery code.

`_refresh_basic_data()` calls `get_base_devices()` and then `_init_default_device()` before `set_device_endpoints_data()`. The base-device lookup excludes thermostats and filters out the Alexa apps, so an account without an online supported voice device raises `NoOnlineDevicesError` before these thermostat endpoints are queried.

Can we let thermostat discovery and polling run without requiring a default voice device? Simply moving default-device selection after discovery would also need care: a thermostat should not become the default target for voice commands.

src/aioamazondevices/implementation/sensor.py
```diff
@@ -256,4 +268,29 @@ def _get_device_sensor_state(
                     scale,
                 )
 
+        if thermostat_data:
```
colindembovsky, line 271: This drops the temperature sensor's error state.

If Amazon returns a property-level error for `temperatureSensor`, the existing parser creates a sensor with `value="n/a"` and `error=True`. This block copies `"n/a"` into `thermostat_data["temperature"]`, removes the original sensor, and publishes the combined thermostat sensor with `error=False`.

The climate platform would then receive a nonnumeric current temperature with no indication that the reading failed. Please preserve the failure information and omit the temperature, or expose it as unavailable, rather than copying the error placeholder into a successful result.

src/aioamazondevices/implementation/thermostat.py
```diff
@@ -0,0 +1,211 @@
...
+        response = await self._http_wrapper.response_to_json(
+            raw_resp, "setEndpointFeature"
+        )
+
+        if format_graphql_error(response):
+            raise CannotSetThermostat(
+                f"Failed to {operation_name} for {device.account_name}"
+            )
+
+        arr = response.get(ARRAY_WRAPPER) or []
```
colindembovsky, line 192: Please reject missing mutation results instead of treating them as success.

`response_to_json()` normalizes an empty JSON response to `{}`. Here that becomes an empty `arr`, then `errors=None`, so the setter returns normally despite receiving no mutation result. A missing `setEndpointFeatures` key has the same effect. Explicit `null` values for `data` or `setEndpointFeatures` instead produce an `AttributeError`.

Please validate the response structure and require a mutation result before returning successfully. Missing or null results should raise `CannotSetThermostat`, so the climate platform can report an unconfirmed command rather than silently accepting it.

colindembovsky (changes requested):
@chemelli74 I've added some review comments.

colindembovsky (changes requested):
Fixing comments

PR #1045, opened 2026-08-27
fix(api): skip vocal history fetch when nothing subscribes

stefan-schweiger:
An `EqualizerStateChange` push event is used as a proxy for "somebody spoke" and calls `_handle_eq_event_as_history_proxy()`, which sleeps 2 seconds, refreshes the CSRF and access tokens, and pulls seven days of vocal history. Only afterwards does `_emit_history_event()` check `on_history_event.frozen`.

A consumer that never subscribes to on_history_event therefore pays for the whole round trip on every one of those events, and the result is discarded. Because `http2.py` awaits the push handler inline in the directive stream loop, it also delays the volume, media and item updates queued behind it.

There is a second effect worth noting: every record is logged at `DEBUG` including `transcriptText` and `personFirstName`, so consumers that forward library logs into user-facing diagnostics end up capturing utterances they never subscribed to.

This moves the frozen check ahead of the fetch, matching the guard the other emitters already use. Consumers that do subscribe are unaffected.

Tests cover both paths, the skip test fails without the fix.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved vocal history event handling by avoiding unnecessary history retrieval when no event listeners are present.
  - Vocal history continues to be retrieved and emitted correctly when listeners are subscribed.

- **Tests**
  - Added coverage for both subscribed and unsubscribed vocal history event scenarios.


jamesonuk:
@stefan-schweiger this will obviously only affect non HA usage so might take a little longer to get round to this.    

Out of interest, are you using the library outside of HA?

chemelli74 (approved):
Thank you so much !

stefan-schweiger:
@jamesonuk yes I used it to migrate my Homey integration from the previous JS implementation that constantly had issues with the authentication to python.

https://github.com/stefan-schweiger/dev.schweiger.echo

jamesonuk (approved):

PR #1051, opened 2026-08-31
feat: add support for Echo Glow

chemelli74:
## Summary by CodeRabbit

* **New Features**
  * Added support for discovering and controlling smart-home lights, including Echo Glow devices.
  * Light controls include power, brightness, color, effects, and tap control.
  * Light status, capabilities, and supported effects are included with device information.
  * Do Not Disturb updates now stay synchronized with real-time device events.

* **Tests**
  * Added coverage for light discovery, state handling, color mapping, validation, and controls.
  * Added CLI support for exercising light devices in integration tests.


src/aioamazondevices/const/queries.py
```diff
@@ -53,6 +53,29 @@
...
+        manufacturerName
+        modelName
+        entityId
+        aliases
+        capabilities
+        alexaDeviceIdentifierList
+      }
+    }
+  }
+}
```
jamesonuk, line 76: This is going to return every single smart home device.
We really want to limit this like we do for AQM devices but I suspect light is going to be far too broad.
https://github.com/chemelli74/aioamazondevices/blob/main/src/aioamazondevices/const/queries.py#L17-L21

Even if filtered to just lights this will bring back HA lights that have been exposed to Alexa and potentially lead to duplicate devices and confusion.

If we can do something so this only brings back the items of interest then it probably should be added to `QUERY_DEVICE_DATA` rather than generating another query.

src/aioamazondevices/structures.py
```diff
@@ -77,6 +100,52 @@ class AmazonDevice:
     notifications: dict[str, AmazonSchedule]
     media_player_supported: bool
     communication_settings: dict[str, str]
+    light: AmazonDeviceLight | None
+
+
+def build_endpoint_device(  # noqa: PLR0913 - a device just has many fields
```
jamesonuk, line 106: This should probably be a separate refactor PR

src/aioamazondevices/const/http.py
```diff
@@ -52,6 +52,7 @@
 URI_MUSIC_PROVIDERS = "api/behaviors/entities"
 URI_NEXUS_GRAPHQL = "nexus/v1/graphql"
 URI_NOTIFICATIONS = "api/notifications"
+URI_PHOENIX_STATE = "api/phoenix/state"
```
jamesonuk, line 55: I have a slight concern here that this is back to the old legacy API.   I know the app still hits this endpoint but equally the other phoenix endpoint was removed which could break things.    Is the data available anywhere else?

src/aioamazondevices/implementation/light.py
```diff
@@ -0,0 +1,473 @@
...
+    Returns ``None`` for anything that is not an Amazon-branded light.
+    """
+    if appliance.get("manufacturerName") != "Amazon":
+        return None
+    if DEVICE_FAMILY_LIGHT not in (appliance.get("applianceTypes") or []):
+        return None
+
+    identifiers = appliance.get("alexaDeviceIdentifierList") or []
+    identity = identifiers[0] if identifiers else {}
+    serial = identity.get("dmsDeviceSerialNumber")
```
jamesonuk, line 138: Is serial not present in the standard `serialNumber` field? (https://github.com/chemelli74/aioamazondevices/blob/d88ea584a0e9472ec0b21916141482016ef7ac02/src/aioamazondevices/const/queries.py#L33) rather than legacy?  (would make incorporating into `QUERY_DEVICE_DATA` easier)

src/aioamazondevices/implementation/light.py
```diff
@@ -0,0 +1,473 @@
...
+            },
+        )
+
+    def _control_for(self, device: AmazonDevice) -> _LightControl:
+        """Return the control data for a known light."""
+        if not (control := self._controls.get(device.serial_number)):
+            raise ValueError(f"Unknown light: {device.serial_number}")
+        return control
+
+    async def _control(self, entity_id: str, operation: dict[str, Any]) -> None:
```
jamesonuk, line 437: Should this not be in https://github.com/chemelli74/aioamazondevices/blob/d88ea584a0e9472ec0b21916141482016ef7ac02/src/aioamazondevices/implementation/sequence.py#L262 like all the other sequence code?

src/aioamazondevices/const/queries.py
chemelli74: Can we filter by vendor ? Currently I get all LIGHTS and then remove all non-Amazon ones.

src/aioamazondevices/const/http.py
chemelli74: I don't know, this is the only endpoint I found for Echo Glow

src/aioamazondevices/implementation/light.py
chemelli74: Will check

src/aioamazondevices/implementation/light.py
chemelli74: take care that `_control` in private not public ;-)

src/aioamazondevices/const/http.py
mmstano: FWIW, in my tests, that was the only endpoint I got as well. My earlier research on the Echo Glow led me to AMP's code which also used that endpoint for the Echo Glow.

src/aioamazondevices/structures.py
chemelli74: Done

PR #1062, opened 2026-09-06
chore: bump browser user agent

### LINKED ISSUE #1018: Persistent HTTP 503 from alexa-comms-mobile-service preferences endpoint — notify Speak/Announce entities never created
# Persistent HTTP 503 from `alexa-comms-mobile-service` preferences endpoint — notify Speak/Announce entities never created

## Describe the bug

The device communications-preferences request returns **HTTP 503 with a `text/html` body** on every poll, and has done so continuously for over 48 hours. Every other Amazon endpoint in the same polling cycle returns 200.

Failing request:

```
GET https://alexa-comms-mobile-service.amazon.com/devicesTypes/A4ZXE0RM7LQ7A/deviceId/<SERIAL>/preferences?devicePreferences=communications&devicePreferences=dropin&devicePreferences=announcements
```

The library retries twice (2s, then 5s) and then falls back with:

```
WARNING [aioamazondevices] Failed to refresh communications settings for device <NAME>, used cached values.
```

**479 consecutive failures** between `2026-08-05 14:20:54` and `2026-08-07 10:06:53`, at the default 5-minute poll interval. Not one success in that window.

The `text/html` content type on a 503 suggests the request is being rejected at an edge/gateway layer rather than by the application itself. Worth noting the library sends `User-Agent: AmazonWebView/AmazonAlexa/2.2.663733.0/iOS/18.5/iPhone`, and Amazon's GraphQL layer already has an explicit `UNOFFICIAL_CLIENT_TOO_MANY_REQUESTS` classification — so unofficial clients are identified and treated differently. This may be a deliberate block or a retirement of this legacy REST path.

## Impact

While cached values existed, the failure was invisible — entities kept working. After the integration was deleted and re-added (wiping the cache), there was nothing to fall back to, and:

- The **Announce** and **Speak** notify entities are **never created at all**
- The **Communications**, **Announcements**, and **Drop In** entities are permanently unavailable
- Setup completes *successfully* with no error surfaced to the user

Combined with home-assistant/core#177402 ("Allow Alexa Devices setup to continue when additional feature APIs fail", shipped in 2026.8), this fails silently: no repair issue, no config-entry error. Automations calling `notify.send_message` against the announce entity break with an unknown-entity error and the cause is not discoverable without debug logs.

Media player, temperature, illuminance, DND, and connectivity entities are all unaffected.

## To Reproduce

1. Set up the Alexa Devices integration with an Echo Dot (5th gen), deviceType `A4ZXE0RM7LQ7A`
2. Enable debug logging for `aioamazondevices`
3. Wait for a poll cycle
4. Observe the 503 on the preferences URL above and the subsequent warning

## Expected behavior

Two separate things:

1. The preferences request should succeed — or, if this endpoint is being retired/blocked for unofficial clients, the preferences should be read from the GraphQL surface at `alexa.amazon.com/nexus/v1/graphql`, which returns 200 throughout and is what the current Alexa app appears to use.
2. Failing that, a persistent failure of this call should not silently drop the notify entities. Either create them optimistically and let the send fail, or raise a repair issue so the user knows announcements are unavailable and why.

## Verified not to be the cause

- **Amazon-side settings are correct.** In the Alexa app, for this device: Communications = Enabled, Announcements = Enabled, Drop In = On.
- **Not the HA entity registry.** Full delete of the config entry → HA restart → clean re-add with fresh authenticator OTP. Same result.
- **Not authentication.** Login succeeds; all other endpoints return 200 with the same session.
- **Not rate limiting.** Rate limits return 429 / `UNOFFICIAL_CLIENT_TOO_MANY_REQUESTS`. This is a 503 that never recovers across 479 attempts and two days.

## System details

- Home Assistant Core: 2026.8.0
- Installation type: Home Assistant OS (on Proxmox)
- aioamazondevices version: 14.2.2
- Device: Amazon Echo Dot (5th gen), deviceType `A4ZXE0RM7LQ7A`
- 2SV via authenticator app: yes
- Last known working: 2026-08-05, before 14:20 local

## Debug logs

```
2026-08-07 10:06:45.810 DEBUG [aioamazondevices] GET request: https://alexa-comms-mobile-service.amazon.com/devicesTypes/A4ZXE0RM7LQ7A/deviceId/<SERIAL>/preferences?devicePreferences=communications&devicePreferences=dropin&devicePreferences=announcements with payload None [json=False]
2026-08-07 10:06:46.126 INFO  [aioamazondevices] Sleeping for 2 seconds before retrying API call to <same URL>
2026-08-07 10:06:48.201 INFO  [aioamazondevices] Sleeping for 5 seconds before retrying API call to <same URL>
2026-08-07 10:06:53.275 DEBUG [aioamazondevices] Response for url <same URL> :
status      : 503
content type: text/html
2026-08-07 10:06:53.275 WARNING [aioamazondevices] Failed to refresh communications settings for device <NAME>, used cached values.
```

For contrast, the same cycle:

```
https://alexa.amazon.com/nexus/v1/graphql          → 200 (endpoint list)
https://alexa.amazon.com/api/dnd/device-status-list → 200
https://alexa.amazon.com/api/notifications          → 200
https://alexa.amazon.com/nexus/v1/graphql           → 200 (sensor state)
```

### LINKED ISSUE #1049: Continuous "Failed to refresh communications settings... used cached values" warnings for all Echo devices (survives re-authentication)
Environment

Home Assistant Core: 2026.8.3
Home Assistant OS: 18.2 (Supervisor 2026.08.0)
Installation type: Home Assistant OS, aarch64, Docker
Python: 3.14.6
Integration: Alexa Devices (built-in, chemelli74)
aioamazondevices version: 14.2.2
Number of Amazon/Alexa devices on the account: ~23

The problem

The integration itself loads successfully (state: loaded, no repair/reauth flag), but the log fills continuously with per-device warnings like:

WARNING (MainThread) [aioamazondevices] Failed to refresh communications settings for device <DeviceName>, used cached values.

This fires for effectively every device on the account, one after another in the same polling pass (seconds apart), and repeats on every refresh cycle — roughly every 8–10 minutes per device, continuously, all day. Example from today's log (times local, IST):

2026-08-29 23:32:14.383 WARNING (MainThread) [aioamazondevices] Failed to refresh communications settings for device Alexa Show Stilt, used cached values.
2026-08-29 23:32:22.267 WARNING (MainThread) [aioamazondevices] Failed to refresh communications settings for device Alexa Show Main Lounge, used cached values.
2026-08-29 23:32:30.169 WARNING (MainThread) [aioamazondevices] Failed to refresh communications settings for device Alexa Basement Bar Show, used cached values.
2026-08-29 23:32:38.040 WARNING (MainThread) [aioamazondevices] Failed to refresh communications settings for device Ekta's Fire TV, used cached values.
2026-08-29 23:32:45.912 WARNING (MainThread) [aioamazondevices] Failed to refresh communications settings for device Alexa Informal Lounge, used cached values.
2026-08-29 23:32:54.477 WARNING (MainThread) [aioamazondevices] Failed to refresh communications settings for device Sam Guest Room Two, used cached values.
2026-08-29 23:33:02.347 WARNING (MainThread) [aioamazondevices] Failed to refresh communications settings for device Alexa Terrace Bar Show, used cached values.
2026-08-29 23:33:10.219 WARNING (MainThread) [aioamazondevices] Failed to refresh communications settings for device Master Bedroom Bose 900, used cached values.

Over a ~15-hour window today, this logged roughly 2,000 times across ~23 devices (~90-110 occurrences per device).

What I've tried

Fully reconfigured/re-authenticated the integration (Settings → Devices & Services → Alexa Devices → Reconfigure) with the account password re-entered. The config entry timestamp confirms the reconfigure succeeded (state: loaded, no errors). The warnings continued immediately afterward, unchanged in frequency, which suggests this isn't a stale-session/cookie issue.
Confirmed the integration is otherwise healthy — no HA repair item, no reauth prompt, entities continue to report values (from cache).

Impact
Low/no functional impact so far — the fallback-to-cache behavior means device state, voice control, and routines appear unaffected. The main issues are (a) log noise at this volume, and (b) the "communications settings" data (drop-in/announcement config) served to HA is presumably stale rather than live.

Suspected cause
Given the account has ~23 devices all polled on the same cycle and this hits an undocumented Amazon endpoint, this looks like Amazon-side rate-limiting on this specific call (similar in shape to other reported 429/rate-limit issues against this integration), rather than an auth problem — but I don't have visibility into the underlying HTTP response since this is logged as a caught warning rather than a raised exception with a status code.

Question for maintainers
Would it be possible to (a) surface the underlying HTTP status/response when this call fails, and/or (b) stagger or de-duplicate this specific per-device call so it doesn't hit ~23 devices back-to-back on every cycle?

francescolf:
Fixes #1049
Fixes #1018 

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the browser identification used for requests to reflect a newer Chrome/Edge version, improving compatibility with services that require current browser signatures.


chemelli74 (approved):

jamesonuk (approved):
My experience is that Amazon don't really take any notice of user agent however this is a safe change

PR #1065, opened 2026-09-07
feat(history): expose recognised speaker from personsInfo

BoBiene:
## What

Alexa's vocal history payload identifies which household member spoke, via the
voice profile that matched the utterance. `alexaHistoryRecords` entries carry a
`personsInfo` object alongside `deviceInfo`:

```json
"personsInfo": {
  "personId": "amzn1.actor.person.oid.REDACTED",
  "personFirstName": "REDACTED",
  "personType": "ADULT"
}
```

`get_vocal_history()` reads five fields out of each record and drops the rest,
so this never reaches consumers. `AmazonVocalRecord` is currently
`timestamp, history_type, intent, title, sub_title`, so there is no way to tell
who gave a command.

This PR adds `person_first_name` and `person_type` to
`AmazonVocalRecord` and populates them from `personsInfo`.

## Why

Home Assistant surfaces vocal history as event entities with `voice_command` /
`voice_reply` / `intent` attributes. Knowing *who* spoke makes per-person
automations possible ("if a child asks for the lights after 21:00…") without
going through a custom skill and the Person Profile API, because the data is already in
the response we fetch.

## How it behaves

- Both fields default to `None`, so existing consumers are unaffected and
  no constructor call needs updating.
- When Alexa does not match a voice profile, `personsInfo` is absent and the
  fields stay `None`.
- `personId` is deliberately not exposed. It is an account-scoped identifier,
  and first names are unique within an Amazon household, so the name is enough
  to tell speakers apart (see review discussion).
- `personsInfo` is accepted as either an object or a single-element list,
  mirroring the handling already applied to `deviceInfo` a few lines above.

## Evidence

Measured against a live account (Echo Dot 4th/5th gen, DE locale) over 31
history records: 24 carried a `personsInfo` block, 7 did not. The four
household voice profiles were each resolved correctly. Names and person IDs are
redacted here and in the test fixture.

## Testing

New `tests/test_history.py` covers the recognised case, the list-shaped
variant, an absent `personsInfo`, an empty list and a payload without the
key at all, and checks that the person ID is not exposed.

```
25 passed
ruff check: All checks passed
ruff format --check: 42 files already formatted
mypy (changed files): Success: no issues found in 3 source files
```

`uv` was unavailable in my environment, so the suite ran in a plain venv with
the pinned tool versions from `.pre-commit-config.yaml` (ruff 0.16.6) rather
than via `uv run pytest`. Note that pre-existing mypy findings in `login.py`
and `http_wrapper.py` are untouched by this change.

## Side note, not part of this PR

`history.py` logs the whole record at debug level:

```python
_LOGGER.debug("Processing vocal history record: %s", record)
```

With this payload that writes the recognised speaker's **first name in clear
text**, once per utterance, into the consumer's log. Given #1056 added further
redactions to debug logging, `scrub_fields()` may be worth applying here too.
Happy to send that as a separate PR if you'd like.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Vocal history records now include recognized speaker details from Alexa, including the speaker’s first name and profile type.
  * Speaker information is supported when provided as either a single entry or a list.

* **Bug Fixes**
  * Missing, empty, or malformed speaker information is handled gracefully, with unavailable details returned as blank values.
  * Account-scoped person IDs are no longer exposed in vocal history records.

### DIFF
--- a/src/aioamazondevices/implementation/history.py
+++ b/src/aioamazondevices/implementation/history.py
@@ -108,12 +108,20 @@ class AmazonHistoryHandler:
                 continue
             serial = device_info["deviceSerialNumber"]
             timestamp = record["timestamp"]
+            person_info = record.get("personsInfo")
+            if isinstance(person_info, list):
+                person_info = person_info[0] if person_info else None
+            if not isinstance(person_info, dict):
+                person_info = {}
             new_record = AmazonVocalRecord(
                 timestamp=timestamp,
                 history_type=utterance_type or record.get("recordType") or "Unknown",
                 intent=record.get("intent") or "Unknown",
                 title=record["title"],
                 sub_title=record["subTitle"],
+                person_id=person_info.get("personId"),
+                person_first_name=person_info.get("personFirstName"),
+                person_type=person_info.get("personType"),
             )
             # Store only the latest record per serial number
             if serial not in records or timestamp > records[serial].timestamp:
--- a/src/aioamazondevices/structures.py
+++ b/src/aioamazondevices/structures.py
@@ -186,6 +186,10 @@ class AmazonVocalRecord:
     intent: str
     title: str
     sub_title: str
+    # Speaker recognised by an Alexa voice profile, when Amazon provides one
+    person_id: str | None = None
+    person_first_name: str | None = None
+    person_type: str | None = None
 
 
 class AmazonListType(StrEnum):
--- a/tests/test_history.py
+++ b/tests/test_history.py
@@ -1,0 +1,76 @@
+# Copyright 2024 Simone Chemelli and contributors
+# SPDX-License-Identifier: Apache-2.0
+
+"""Tests for Alexa vocal history parsing."""
+
+from typing import Any
+from unittest.mock import AsyncMock
+
+import pytest
+
+from aioamazondevices.implementation import history as history_module
+from aioamazondevices.implementation.history import AmazonHistoryHandler
+
+from .const import TEST_SERIAL_1
+
+PersonsInfo = dict[str, str] | list[dict[str, str]] | None
+
+TEST_PERSON = {
+    "personId": "amzn1.actor.person.oid.PERSON_ID",
+    "personFirstName": "Alice",
+    "personType": "ADULT",
+}
+
+
+def _record(persons_info: PersonsInfo) -> dict[str, Any]:
+    """Build a minimal vocal history record, shaped like the Amazon payload."""
+    return {
+        "timestamp": 1757000000000,
+        "utteranceType": "GENERAL",
+        "intent": "PlayMusicIntent",
+        "title": "play some music",
+        "subTitle": "Echo Dot",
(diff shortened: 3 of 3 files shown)


src/aioamazondevices/implementation/history.py
```diff
@@ -108,12 +108,20 @@ async def get_vocal_history(self) -> dict[str, AmazonVocalRecord]:
...
+                person_info = person_info[0] if person_info else None
+            if not isinstance(person_info, dict):
+                person_info = {}
             new_record = AmazonVocalRecord(
                 timestamp=timestamp,
                 history_type=utterance_type or record.get("recordType") or "Unknown",
                 intent=record.get("intent") or "Unknown",
                 title=record["title"],
                 sub_title=record["subTitle"],
+                person_id=person_info.get("personId"),
```
jamesonuk: Only real question is whether this is useful?
It is something like `amzn1.actor.person.oid.<xxx>` so not sure of it's value

BoBiene: Fair question. On its own the ID isn't human-readable, but it's the only stable key for the speaker:

- `personFirstName` is a display name, not an identifier. It isn't guaranteed to be unique within a household (two members can share a first name), so anything keyed on it can pick the wrong person. The ID doesn't have that problem.
- It lets consumers tell speakers apart without persisting a first name, e.g. map the ID to a person once in their own config and keep names out of state history and logs. That ties in with the logging note in the description.

That said, if you'd rather keep the record lean, I'm happy to drop `person_id` and keep only `person_first_name` and `person_type`. Just let me know.

jamesonuk: Just gone in to Amazon and tried to change my name to match the wife's and it would not let me.   I think Amazon will enforce first name being unique (Checked with kids too)
image

I would prefer not to store the ID because it is unique to your account and could be used to access your account at some level rather than keeping the number of fields down.

Unless there is a good reason why it is needed then we should remove it.

BoBiene: Thanks for checking, that settles it. I've removed `person_id` in f74bdf5, so the record now only carries `person_first_name` and `person_type`, and I've updated the PR description accordingly.

chemelli74 (approved):
LGTM !

jamesonuk:
> > > In `@tests/test_history.py` around lines 46 - 61, Add a test case for the
> > > `_record` input with the personsInfo key omitted entirely, and assert it
> > > produces `(None, None)` for the exposed fields. Keep the existing None and
> > > empty-list cases unchanged, using the existing parametrized test structure.

@BoBiene could you just look at this which I think is a valid issue then we should be good to go

BoBiene:
Good catch, thanks. Added in c85125f: the parametrized test now also covers a payload that carries no `personsInfo` key at all, which is what Amazon sends when no voice profile matched. The existing None and empty-list cases are unchanged. 25 passed, ruff and mypy clean.

jamesonuk (approved):

PR #1079, opened 2026-09-13
fix: preserve item name casing in get_list_items

### LINKED ISSUE #1078: get_list_items() uses str.capitalize(), which lowercases everything after the first character
### Description

In `AmazonToDoHandler.get_list_items()` (src/aioamazondevices/implementation/todo.py), 
each item's name is transformed with Python's `str.capitalize()`:

    name=(item_info["itemName"]).capitalize(),

`str.capitalize()` doesn't just uppercase the first character — it also forces every 
other character in the string to lowercase. This silently destroys any casing already 
present in the stored item name, on every read (which also means every time the entity 
state / todo.get_items response is refreshed).

### Reproduction

    >>> "@fruits et légumes@ Champignon".capitalize()
    '@fruits et légumes@ champignon'
    >>> "Riz tang frère".capitalize()
    'Riz tang frère'   # looks fine here...
    >>> "@tang frères@ Riz tang frère".capitalize()
    '@tang frères@ riz tang frère'   # ...but breaks as soon as anything precedes it

This is especially disruptive for any consumer that prefixes item names (e.g. Home 
Assistant cards that store a category as a "@category@ name" prefix in the item name, 
since the todo entity has no separate category field) — the actual item name never 
keeps its casing, no matter what is sent to add_item()/rename_item(), because it gets 
re-mangled on every subsequent fetch.

### Expected behavior

Preserve the item name's casing as stored, or at most capitalize only the very first 
alphabetic character without lowercasing the rest, e.g.:

    def _cap_first(s: str) -> str:
        return s[:1].upper() + s[1:] if s else s

    ...
    name=_cap_first(item_info["itemName"]),

### Environment

- aioamazondevices (installed via the Home Assistant "Alexa Devices" core integration)
- Reported September 2026

Hkain77:
Fixes #1078

## Summary by CodeRabbit

- **Bug Fixes**
  - Item names now capitalize only the first character while preserving the capitalization of the remaining characters.
  - Empty item names remain unchanged.


chemelli74:
Hi @Hkain77, thanks for your contribution!

However, we don’t want to manipulate the data.
We simply capitalize the first letter, as Home Assistant does in many places, and leave everything else unchanged.

Hkain77:
> Hi @Hkain77, thanks for your contribution!
> 
> However, we don’t want to manipulate the data. We simply capitalize the first letter, as Home Assistant does in many places, and leave everything else unchanged.

Hi @chemelli74, thanks for looking at this!

I think there might be a misunderstanding about what str.capitalize() actually does — 
it doesn't just capitalize the first letter, it also forces every other character in 
the string to lowercase. So it does modify the rest of the data, it just isn't obvious 
unless the name has more than one capital letter or word.

You can check it directly in a Python shell:

    >>> "Riz Tang Frère".capitalize()
    'Riz tang frère'

The "F" in "Frère" gets silently lowercased, even though nothing in that name should 
have been touched. This also breaks any caller (like Home Assistant's "@category@ name" 
prefix pattern used in some shopping-list Lovelace cards) where the actual item name 
doesn't start at position 0 of the string — the whole name after the prefix gets 
flattened to lowercase.

The fix I proposed only touches the very first character (str[:1].upper() + str[1:]) 
and genuinely leaves everything else byte-for-byte unchanged, which I believe was the 
original intent.

Happy to add a quick unit test demonstrating this if that helps move it forward.

PR #1080, opened 2026-09-13
feat: expose sensor timestamps

chemelli74:
Add `AmazonDeviceSensor.time_of_sample`, populated from the timeOfSample
already requested by the getEndpointState query, so downstream consumers
can detect stale samples that Amazon's cloud keeps re-serving.

Fixes part of home-assistant/core#182083

## Summary by CodeRabbit

* **New Features**
  * Sensor data now includes the time each reading was sampled when available.
  * Valid timestamps are converted to timezone-aware date and time values for consistent display and processing.

* **Bug Fixes**
  * Invalid or unsupported sample timestamps no longer cause errors; they are safely ignored and logged.
  * Missing sample timestamps continue to be handled gracefully, leaving the sample time unavailable rather than affecting sensor data.


src/aioamazondevices/implementation/communication.py
```diff
@@ -94,10 +94,11 @@ async def get_communication_preferences(
                 _, resp = await self._http_wrapper.session_request(
                     method=HTTPMethod.GET, url=url
                 )
-            except CannotRetrieveData:
+            except CannotRetrieveData as err:
                 _LOGGER.warning(
-                    "Failed to refresh communications settings for device %s, used cached values.",  # noqa: E501
+                    "Failed to refresh communications settings for device %s, used cached values: %s",  # noqa: E501
                     device.account_name,
+                    err,
```
jamesonuk: This is more related to this particular log entry rather than the the point of the PR.
Given this is logging many, many times (I am up to 443 and I restarted only a few hours ago)

Image

Do we want to add the exception here ??

jamesonuk (commented):
The time of sample part looks solid and worthwhile.
The only question I have is whether we actually want to log the exception for the communication errors

src/aioamazondevices/implementation/communication.py
chemelli74: I think we should log (you can eventually filter in logger config on HA side), and yes adding at least the HTTP status or main reason would be good.

Separate PR is better probably.
I'l remove the logging part from this one

jamesonuk (approved):

src/aioamazondevices/implementation/sensor.py
```diff
@@ -25,6 +26,29 @@
 from aioamazondevices.utils import _LOGGER, format_graphql_error
 
 
+def _parse_sample_timestamp(
+    feature_property: dict[str, Any],
```
jamesonuk: I would prefer to pass in the raw_timestamp rather than the whole `dict` here

PR #1082, opened 2026-09-14
refactor(device): extract device creation into a helper

chemelli74:


jamesonuk (approved):
Tested now and AQM device still works with no regression

PR #1089, opened 2026-09-17
feat: support optional display text for announcements

uholland:
Allow callers to provide an optional `display_text` when sending an announcement.

This lets the screen show normal spelling and version numbers while the spoken message uses pronunciation-friendly text.

```python
await api.call_alexa_announcement(
    device,
    "Version one thirty-two zero is ready.",
    display_text="Version 1.32.0 is ready.",
)
```

## Compatibility

- Omitting `display_text`, or passing `None`, preserves existing behavior.
- The spoken message, display title, and target devices remain unchanged.
- Announcement batching preserves different display texts.
- This change does not add SSML support.

## Validation

- All 45 tests pass, including eight new cases.
- All project pre-commit checks pass.
- A live Echo test confirmed the requested pronunciation and separate display text, including a dotted version number.
- The installed Home Assistant integration and library were not modified for this test.

## Summary by CodeRabbit

* **New Features**
  * Added an optional display text setting for announcements, allowing the on-screen message to differ from the spoken message.
  * Existing announcement behavior remains unchanged when display text is omitted or set to `None`.
  * Supports empty display text while preserving the announcement title and speech content.

* **Documentation**
  * Added usage guidance describing the display text option and its behavior.


jamesonuk:
Can I just check what device(s) this is actually for as obviously most don't have any display.
Also what happens if you send display text to a standard echo without a display?

uholland:
@jamesonuk I tested this on an Echo Show 5. The screen showed the separate display text, while speech used the original message.

Announcements already include a `display` object. This change only allows overriding its body; the speech payload stays unchanged. I therefore expect screenless Echos to speak normally, but I have not verified that on a physical device without a display.

PR #1090, opened 2026-09-17
feat: support opt-in SSML for announcements

uholland:
## Summary

Add an optional `speech_type` argument to announcements. It defaults to
`"text"` and accepts `"ssml"` for formatted speech.

This allows callers to control pronunciation, pauses, emphasis, and speech
rate while keeping readable text on the device screen.

## Example

```python
await api.call_alexa_announcement(
    device,
    '<speak><sub alias="Example">EXAMPLE</sub>'
    '<break time="500ms"/> is ready.</speak>',
    display_text="EXAMPLE is ready.",
    speech_type="ssml",
)
```

## Behavior and compatibility

- Existing calls continue to use plain text.
- SSML requires an explicit `display_text`.
- The speech payload reaches Amazon without modification.
- Invalid speech types and malformed XML are rejected before queuing.
- Validation requires one `speak` root and supports Amazon-prefixed tags.
- DTD and entity declarations are rejected.
- Batching keeps plain-text and SSML announcements separate.
- The change does not alter persistent device settings.

XML validation does not guarantee that Amazon supports every tag or
combination.

## Validation

- All 56 tests pass, including 11 additional SSML cases.
- All project pre-commit checks pass.
- Live tests on an Echo device with German speech confirmed:
  - Pronunciation aliases and IPA phonemes.
  - Pauses and speech-rate changes.
  - Emphasis, pitch, and relative volume.
  - Numbers, spelling, dates, and times.
  - Whispered speech.
- Separate screen text remained readable during the tests.
- The installed Home Assistant integration and library were not modified.

These tests cover the selected examples on one device and locale.
They do not establish support for every SSML feature or Alexa device.

## Dependency

This change builds on the optional announcement display text introduced
in #1089.


jamesonuk:
@uholland I think this already works
https://www.home-assistant.io/integrations/alexa_devices#using-advanced-markup-in-a-notification
just take out the `<speak>` elements

```yaml
action: notify.send_message
data:
  message: <sub alias="Example">EXAMPLE</sub> <break time="500ms"/> is ready
  title: title
target:
  entity_id: notify.study_dot_speak

```

uholland:
@jamesonuk Thanks! I tested Speak without the `<speak>` wrapper. Speech worked, but nothing appeared on the display. The integration does not forward `title`.

I need custom pronunciation and visible text—for example, saying “Doctor Smith” while displaying “Dr. Smith”. This PR adds SSML to announcements, alongside the separate display text in #1089.

Is there an existing way to achieve both?

